PR #9, opened 2016-02-11
- Allow the user agent to be specified when the SDK is used

marksztainbok-okta:
- Expose the item type of an array in the schema models
- Make all the public methods in the clients virtual to allow consumers to easily mock the methods for improved testability
- Fix exception rethrows so that they preserve the original stack trace
- Ran Resharper to perform some code cleanup and remove redundant code

Relates to: OKTA-77612


raphaellondner-okta:
Successfully reviewed and regression tested changes

marksztainbok-okta:
:+1:

PR #18, opened 2016-12-13
Added delete user functionality

raphaellondner-okta:
Added UsersClient.Delete method, Delete-OktaUser PowerShell method and UsersClientTests DeleteUser test method


raphaellondner-okta:
@marksztainbok-okta Would you be available to review this PR, please?

Okta.Core.Automation/Users/DeleteOktaUser.cs
```diff
@@ -0,0 +1,103 @@
...
+namespace Okta.Core.Automation
+{
+    [Cmdlet("Delete", "OktaUser")]
+    public class DeleteOktaUser : OktaCmdlet
+    {
+        [Parameter(
+            Mandatory = false,
+            ValueFromPipelineByPropertyName = true,
+            Position = 0,
+            HelpMessage = "Id to retrieve"
```
marksztainbok-okta: ID

Okta.Core.Tests/Clients/Helpers.cs
```diff
@@ -155,5 +156,36 @@ internal static Models.Group CreateGroup(OktaClient oktaClient, TestGroup dbGrou
...
+            string allowedChars,
+            int minLength,
+          int maxLength,
+            int count,
+            Random rng)
+        {
+            char[] chars = new char[maxLength];
+            int setLength = allowedChars.Length;
+
+            while (count-- > 0)
```
marksztainbok-okta: Could this just be a for loop?

Okta.Core.Tests/Clients/Helpers.cs
```diff
@@ -155,5 +156,36 @@ internal static Models.Group CreateGroup(OktaClient oktaClient, TestGroup dbGrou
...
         }
 
+        internal static string GetRandomString()
+        {
+            const string AllowedChars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz#@$^*()";
+            Random rng = new Random();
+
+            var randomStrings = RandomStrings(AllowedChars, 8, 16, 1, rng);
+
+            return randomStrings.ElementAt<string>(0);
```
marksztainbok-okta: If you just want a single random string here, why not just make RandomStrings return a single string instead of an Enumerator with multiple?

Okta.Core.Tests/Clients/UsersClientTests.cs
```diff
@@ -196,6 +196,37 @@ public void AuthenticateUser()
...
+                if (existingUser != null)
+                {
+                    usersClient.Delete(existingUser);
+                    existingUser = usersClient.GetByUsername(strUserLogin);
+                    Assert.IsNull(existingUser, "Okta User {0} could be retrieved but should have been deleted: {1}", dbUser.Login, strEx);
+                }
+            }
+            catch (OktaException e)
+            {
+                strEx = string.Format("Error Code: {0} - Summary: {1} - Message: {2}", e.ErrorCode, e.ErrorSummary, e.Message);
```
marksztainbok-okta: Should an exception be considered a test failure? If so the exception should be thrown or `Assert.Fail` should be called

Okta.Core.Automation/Properties/AssemblyInfo.cs
```diff
@@ -32,5 +32,6 @@
 // You can specify all the values or you can default the Build and Revision Numbers 
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("0.1.5.4")]
-[assembly: AssemblyFileVersion("0.1.16280.1919")]
+[assembly: AssemblyVersion("2016.12.12.2312")]
```
raphaellondner-okta: Good point. Talking this discussion offline.

Okta.Core.Automation/Properties/AssemblyInfo.cs
```diff
@@ -8,7 +8,7 @@
 [assembly: AssemblyTitle("Okta.Core.Automation")]
 [assembly: AssemblyDescription("")]
 [assembly: AssemblyConfiguration("")]
-[assembly: AssemblyCompany("Okta")]
+[assembly: AssemblyCompany("Okta, Inc")]
 [assembly: AssemblyProduct("Okta PowerShell commandlets")]
 [assembly: AssemblyCopyright("Copyright © Okta 2016")]
```
marksztainbok-okta: Yes

Okta.Core.Tests/Clients/UsersClientTests.cs
raphaellondner-okta: Added Assert.Fail in next commit

raphaellondner-okta:
@marksztainbok-okta Thanks for the code review, here's the [latest commit](https://github.com/okta/oktasdk-csharp/pull/18/commits/ef32f6e4def884e57d616325065bc070c0b7f4fa).

raphaellondner-okta:
@marksztainbok-okta do you think you'll have the time to take a look at my last commit? Thanks :)

Okta.Core.Automation/Users/DeleteOktaUser.cs
```diff
@@ -0,0 +1,103 @@
...
+                        searchType = SearchType.ElasticSearch;
+                    }
+                    var usersEnum = usersClient.GetFilteredEnumerator(query: Query, searchType: searchType, filter: new FilterBuilder(Filter));
+                    users = usersEnum.ToList<User>();
+                }
+
+                foreach (User user in users)
+                {
+                    string strUserLogin = user.Profile.Login;
+                    bool bContinue = false;
```
rchild-okta: nit: why bContinue? How about something like "deactivated"?

Okta.Core.Automation/Users/DeleteOktaUser.cs
```diff
@@ -0,0 +1,103 @@
...
+                    try
+                    {
+                        usersClient.Deactivate(user);
+                        bContinue = true;
+                    }
+                    catch (OktaException oex2)
+                    {
+                        if (oex2.ErrorCode == OktaErrorCodes.ResourceNotFoundException)
+                        {
+                            bContinue = true;
```
rchild-okta: Why do we continue to try to delete the user if they're not found?

Okta.Core.Automation/Users/GetOktaUser.cs
```diff
@@ -22,31 +22,56 @@ public class GetOktaUser : OktaCmdlet
...
+
+                if (!string.IsNullOrEmpty(Id))
+                {
+                    var user = usersClient.Get(Id);
+                    WriteObject(user);
+                }
+                else
+                {
+                    SearchType searchType = SearchType.Filter;
+                    if(Search)
```
rchild-okta: nit: space between if and (

Okta.Core.Tests/Clients/Helpers.cs
```diff
@@ -155,5 +156,31 @@ internal static Models.Group CreateGroup(OktaClient oktaClient, TestGroup dbGrou
...
+            const string AllowedChars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz#@$^*()";
+            Random rng = new Random();
+
+            return RandomString(AllowedChars, 8, 16, rng);
+        }
+
+        private static string RandomString(
+            string allowedChars,
+            int minLength,
+          int maxLength,
```
rchild-okta: nit: spacing

Okta.Core.Tests/Clients/Helpers.cs
```diff
@@ -155,5 +156,31 @@ internal static Models.Group CreateGroup(OktaClient oktaClient, TestGroup dbGrou
...
+        {
+            const string AllowedChars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz#@$^*()";
+            Random rng = new Random();
+
+            return RandomString(AllowedChars, 8, 16, rng);
+        }
+
+        private static string RandomString(
+            string allowedChars,
+            int minLength,
```
rchild-okta: Just curious - why do we need so much variability here? I.e. seems like all we really want is "give me a random string of length x". Not a big deal, but feels like we're overcomplicating this helper function that is only used in one place.

Okta.Core.Tests/Clients/UsersClientTests.cs
```diff
@@ -464,24 +484,40 @@ public void GetUserByUsername()
...
-        //    return tenant;
-        //}
+                if (!string.IsNullOrWhiteSpace(strUserID))
+                {
+                    Models.User existingUser = usersClient.Get(oktaTenant.TestUserId);
+
+                    Assert.IsFalse(existingUser == null, "The user with id {0} doesn't exist", strUserID);
+
+                    Models.LoginCredentials loginCreds = new Models.LoginCredentials();
+                    loginCreds.Password.Value = Helpers.GetRandomString();
```
rchild-okta: And actually in this case, I'm wondering why we're even setting a random string in the first place - since it's for tests, why not always use the same password? And if we get into a situation where we need to test the change password flows, use a different known one there (always making sure to cleanup the test at the end or reset the env to a known state).

Okta.Core.Tests/Clients/UsersClientTests.cs
```diff
@@ -464,24 +484,40 @@ public void GetUserByUsername()
...
+                    Models.User updatedUser = usersClient.SetCredentials(strUserID, loginCreds);
+
+                    Assert.IsNotNull(updatedUser, "The updated user is null so there likely was an error while updating his password.");
+
+                }
+            }
+            catch (OktaException e)
+            {
+                string strEx = string.Format("Error Code: {0} - Summary: {1} - Message: {2}", e.ErrorCode, e.ErrorSummary, e.InnerException.InnerException.Message);
+                //Console.WriteLine(strEx);
```
rchild-okta: Similar comment to @marksztainbok-okta's earlier - should an exception cause a failure? And, remove commented lines.

Okta.Core.Automation/Users/DeleteOktaUser.cs
```diff
@@ -0,0 +1,103 @@
...
+        protected override void ProcessRecord()
+        {
+            var usersClient = Client.GetUsersClient();
+            try
+            {
+                List<User> users = new List<User>();
+                if (!string.IsNullOrEmpty(Id))
+                {
+                    var user = usersClient.Get(Id);
+                    if (user != null)
```
yiyangli-okta: missing `{}`

Okta.Core.Tests/Clients/Helpers.cs
```diff
@@ -155,5 +156,31 @@ internal static Models.Group CreateGroup(OktaClient oktaClient, TestGroup dbGrou
             return oktaGroup;
         }
 
+        internal static string GetRandomString()
+        {
+            const string AllowedChars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz#@$^*()";
```
yiyangli-okta: lower case of `allowedChars` within a method,  unless you want to make it a static field

Okta.Core.Automation/Users/DeleteOktaUser.cs
```diff
@@ -0,0 +1,103 @@
...
+                    {
+                        if (oex2.ErrorCode == OktaErrorCodes.ResourceNotFoundException)
+                        {
+                            bContinue = true;
+                        }
+                    }
+
+                    if (bContinue)
+                    {
+                        usersClient.Delete(user);
```
yiyangli-okta: will `Deactivate` or `Delete` change the reference `user` or its properties? because foreach is supposed to be a read-only iteration.

Okta.Core.Automation/Users/DeleteOktaUser.cs
raphaellondner-okta: good suggestion!

Okta.Core.Tests/Clients/Helpers.cs
raphaellondner-okta: Fair point. Probably an overkill but I was looking for a function that would generate a random string quick and dirty (since it's only used in the context of password change unit test and this code popped up. Fine with simplifying/refactoring this at a later point though.

Okta.Core.Tests/Clients/UsersClientTests.cs
raphaellondner-okta: Done this way specifically because I had no way of resetting the environment to a known state (for instance, without that user in the first place). Now possible with the user delete feature, but previously, this is how I worked around it in this context.

Okta.Core.Automation/Users/DeleteOktaUser.cs
raphaellondner-okta: It will. What do you recommend using in this situation to allow deleting this potential list of users?

raphaellondner-okta: Note that I tested it worked with a real list of users (filtered by username - they all started with "testuser")

raphaellondner-okta:
[Code fixes](https://github.com/okta/oktasdk-csharp/pull/18/commits/506bf4d4b258b2e70b48fe2650f323e4e6cb2aa3) @rchild-okta @yiyangli-okta

Okta.Core.Automation/Users/DeleteOktaUser.cs
raphaellondner-okta: Fixed in [separate commit](https://github.com/okta/oktasdk-csharp/pull/18/commits/77dbc421811bd23e77cf4417bb443ff0d370a070).

Okta.Core.Tests/Clients/UsersClientTests.cs
```diff
@@ -509,13 +509,12 @@ public void SetCredentials()
...
 
                     Assert.IsNotNull(updatedUser, "The updated user is null so there likely was an error while updating his password.");
-
                 }
             }
             catch (OktaException e)
             {
                 string strEx = string.Format("Error Code: {0} - Summary: {1} - Message: {2}", e.ErrorCode, e.ErrorSummary, e.InnerException.InnerException.Message);
-                //Console.WriteLine(strEx);		
+                Console.WriteLine(strEx);
```
rchild-okta: As in @marksztainbok-okta's comment, we shouldn't be writing it, we should be throwing it if we get into a bad state.

PR #19, opened 2016-12-28
Disabling DNS caching

raphaellondner-okta:
Prevents DNS caching - Standard .NET Framework client will still need to set
ServicePointManager.DnsRefreshTimeout = 0 before initializing an Okta .NET client
(no current solution for portable .NET clients)


raphaellondner-okta:
@rchild-okta, @swastimishra-okta and @marksztainbok-okta , could you please review this change? It's a one-liner but potentially with fairly important consequences so your feedback is much appreciated

rchild-okta:
As discussed, let's close this PR for now - 2 minutes as the default sounds reasonable to me, no need to reduce it to 0.

PR #30, opened 2017-03-17
Port to netstandard

nbarbettini:
Progress on porting the SDK to .NET Standard.

(Best to look at the diff with [whitespace ignored](https://github.com/okta/oktasdk-csharp/pull/30/files?w=1)).

The new `Okta.Core.Client.csproj` project targets .NET Standard 1.1, which means it is usable from .NET Framework 4.5 *and* .NET Core projects. This does mean we will need to drop support for Silverlight 5 and .NET Framework 4.0. Since this is potentially a breaking change, I've incremented the major version number to 1.0.0.

Outstanding todos:

* All of the tests that rely on SQL Express will fail currently. We'll need to figure out some other way of doing data-driven tests.
* I think the Automation (PowerShell cmdlet) project should be moved to its own solution and repo. If it needs to rely on the SDK, it should do so through a public NuGet package reference.


src/Okta.Core.Client/RegexUtils.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Globalization;
+
+namespace Okta.Core
+{
+    /// <summary>
+    /// Regular Expressions utilities
+    /// </summary>
+    public class RegexUtilities
```
jmelberg-okta: Thinking we should remove this class. It is only used for unit tests and doesn't need to be in the core project.

src/Okta.Core.Client/OktaJsonConverter.cs
```diff
@@ -0,0 +1,301 @@
...
+
+                                // Add it to our dictionary
+                                linkDictionary.Add(linkJTokenName, linkList);
+
+                            }
+
+                            v = linkDictionary;
+                        }
+
+                        //2016-01-29, RL, bug fix to handle the Hints dictionary case
```
jmelberg-okta: We have massive commented code. We should clean this up

src/Okta.Core.Client/Models/Users/Provider.cs
```diff
@@ -0,0 +1,16 @@
...
+    /// </summary>
+    public class Provider : ApiObject
+    {
+        [JsonProperty("type")]
+        public string Type { get; set; }
+
+        [JsonProperty("name")]
+        public string Name { get; set; }
+    }
+}
```
jmelberg-okta: nit: Newline

src/Okta.Core.Client/Models/MFA/FactorProviderType.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// </summary>
+    public static class FactorProviderType
+    {
+        public const string Okta = "OKTA";
+        public const string Symantec = "SYMANTEC";
+        public const string Google = "GOOGLE";
+        public const string Duo = "DUO";
+        public const string RSA = "RSA";
+        public const string Yubico = "YUBICO";
+
```
jmelberg-okta: nit: Extra line

src/Okta.Core.Client/Models/Apps/App.cs
```diff
@@ -0,0 +1,305 @@
...
+                OptionalField2 = optionalField2, 
+                OptionalField2Value = optionalField2Value, 
+                OptionalField3 = optionalField3, 
+                OptionalField3Value = optionalField3Value
+            };
+
+            return app;
+        }
+
+        /*// Build SAML 2.0
```
jmelberg-okta: @lboyette-okta: Any idea why we have this comment block?

lboyette-okta: I don't think I got it working. Removing it's the right thing to do

nbarbettini:
I'm not going to spend any time on this, since it would have been a breaking change and is superseded by 1.x. 0.x won't be compatible with netstandard unless we decide to revisit this.

PR #36, opened 2017-04-12
:bug: Fixes GetUser call in UsersClient

jmelberg-okta:
:bug: Replaces the way we query for users

Resolves: OKTA-119500

### DIFF
--- a/Okta.Core/Clients/UsersClient.cs
+++ b/Okta.Core/Clients/UsersClient.cs
@@ -51,18 +51,30 @@
         /// <example>userClient.GetByUsername("user@domain.local") or userClient.GetByUsername("user")</example>
         public virtual User GetByUsername(string userName)
         {
-            User user = null;
+            if (!userName.StartsWith("?"))
+            {
+                return base.Get(userName);
+            }
+
             var filter = new FilterBuilder();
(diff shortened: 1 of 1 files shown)


Okta.Core/Clients/UsersClient.cs
```diff
@@ -51,6 +51,11 @@ public virtual User Get(string userId)
...
         public virtual User GetByUsername(string userName)
         {
+            if (!userName.StartsWith("?"))
+            {
+                return base.Get(userName);
+            }
+
             User user = null;
             var filter = new FilterBuilder();
             filter.Where("profile.login").EqualTo(userName);
```
lboyette-okta: Instead of  `return user;`, let's actually do `return base.Get(user.Id);`

This will do two requests, but we'll be guaranteed the links.

Okta.Core/Clients/UsersClient.cs
```diff
@@ -62,7 +67,7 @@ public virtual User GetByUsername(string userName)
                 user = usersEnum.Current;
             }
             
-            return user;
+            return base.Get(user.Id);
```
lboyette-okta: This will throw an error if the user doesn't exist

jmelberg-okta:
Error:

PR #39, opened 2017-04-24
Add XFF Header as part of authentication

jyotsnaraghunathan-okta:
Client needs to pass XFF Header in the authn request to resolve network
zone based MFA policies.

### DIFF
--- a/Okta.Core/Clients/AuthClient.cs
+++ b/Okta.Core/Clients/AuthClient.cs
@@ -60,50 +60,6 @@
             return Utils.Deserialize<AuthResponse>(response);
         }
 
-
-        /// <summary>
-        /// Authenticates an Okta user with Context Object
-        /// </summary>
-        /// <param name="username">User's username/login</param>
-        /// <param name="password">User's password</param>
-        /// <param name="deviceToken">User's device token</param>
-        /// <param name="ipAddress">User's IP address</param>
-        /// <param name="userAgent">User's userAgent</param>
-        /// <param name="relayState">opaque value for the transaction and processed as untrusted data which is just echoed in a response. It is the client’s responsibility to escape/encode this value before displaying in a UI such as a HTML document </param>
-        /// <param name="bWarnPasswordExpired">Optional parameter indicating whether the PASSWORD_WARN status should be returned if available. Defaults to false</param>
-        /// <param name="bMultiOptionalFactorEnroll">Optional parameter indicating whether the user should be prompted to add an additional second factor if available </param>
-        /// <returns></returns>
-        public virtual AuthResponse AuthenticateWithContext(string username, string password, string deviceToken, string ipAddress = null, string userAgent = null, string relayState = null, bool bWarnPasswordExpired = false, bool bMultiOptionalFactorEnroll = false)
-        {
-            bool bAddXForwardedHeader = false;
-
-            var authRequest = new AuthRequest
-            {
-                Username = username,
-                Password = password,
-                RelayState = relayState,
-                Options =
-                {
-                    WarnBeforePasswordExpiration = bWarnPasswordExpired,
-                    MultiOptionalFactorEnroll = bMultiOptionalFactorEnroll
-                }
-            };
-
-            if (!string.IsNullOrEmpty(ipAddress.Trim()))
-            {
-                authRequest.Context = new AuthContext { IpAddress = ipAddress.Trim() };
-                bAddXForwardedHeader = true;
-            }
-            if (!string.IsNullOrEmpty(deviceToken.Trim()))
-                authRequest.Context.DeviceToken = deviceToken;
-
-            if (!string.IsNullOrEmpty(userAgent.Trim()))
-                authRequest.Context.UserAgent = userAgent;
-
-            var response = BaseClient.Post(resourcePath, authRequest.ToJson(), true, bAddXForwardedHeader);
-            return Utils.Deserialize<AuthResponse>(response);
-        }
-
         public virtual AuthResponse Enroll(string stateToken, Factor factor)
         {
             factor.SetProperty("stateToken", stateToken);
--- a/Okta.Core/Clients/AuthClient.cs
+++ b/Okta.Core/Clients/AuthClient.cs
@@ -42,6 +42,49 @@
             return Utils.Deserialize<AuthResponse>(response);
         }
 
+        /// <summary>
+        /// Authenticates an Okta user with Context Object
+        /// </summary>
+        /// <param name="username">User's username/login</param>
+        /// <param name="password">User's password</param>
+        /// <param name="deviceToken">User's device token</param>
+        /// <param name="ipAddress">User's IP address</param>
+        /// <param name="userAgent">User's userAgent</param>
+        /// <param name="relayState">opaque value for the transaction and processed as untrusted data which is just echoed in a response. It is the client’s responsibility to escape/encode this value before displaying in a UI such as a HTML document </param>
+        /// <param name="bWarnPasswordExpired">Optional parameter indicating whether the PASSWORD_WARN status should be returned if available. Defaults to false</param>
+        /// <param name="bMultiOptionalFactorEnroll">Optional parameter indicating whether the user should be prompted to add an additional second factor if available </param>
+        /// <returns></returns>
+        public virtual AuthResponse AuthenticateWithContext(string username, string password, string deviceToken, string ipAddress = null, string userAgent = null, string relayState = null, bool bWarnPasswordExpired = false, bool bMultiOptionalFactorEnroll = false)
+        {
+            bool bAddXForwardedHeader = false;
+
+            var authRequest = new AuthRequest
+            {
+                Username = username,
+                Password = password,
+                RelayState = relayState,
+                Options =
+                {
+                    WarnBeforePasswordExpiration = bWarnPasswordExpired,
+                    MultiOptionalFactorEnroll = bMultiOptionalFactorEnroll
+                }
+            };
+
+            if (!string.IsNullOrEmpty(ipAddress.Trim()))
+            {
+                authRequest.Context = new AuthContext { IpAddress = ipAddress.Trim() };
+                bAddXForwardedHeader = true;
+            }
+            if (!string.IsNullOrEmpty(deviceToken.Trim()))
+                authRequest.Context.DeviceToken = deviceToken;
+
+            if (!string.IsNullOrEmpty(userAgent.Trim()))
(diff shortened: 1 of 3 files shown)


nbarbettini:
@jyotsnaraghunathan-okta I want to make sure I'm understanding the use case here:

1. The developer's application receives a request from a user, and needs to make a trusted request to Okta via the Okta SDK to fulfill it.
2. The Okta policy is IP-based and so the *client's* IP address needs to be sent to Okta using X-Forwarded-For.
3. When the application calls the Okta SDK, it explicitly passes the client IP from the incoming request to the SDK method.

Is that the use case you're solving for?

jyotsnaraghunathan-okta:
Yes. That's correct.

On Apr 24, 2017, at 5:58 PM, Nate Barbettini <notifications@github.com<mailto:notifications@github.com>> wrote:


@jyotsnaraghunathan-okta<https://github.com/jyotsnaraghunathan-okta> I want to make sure I'm understanding the use case here:

  1.  The developer's application receives a request from a user, and needs to make a trusted request to Okta via the Okta SDK to fulfill it.
  2.  The Okta policy is IP-based and so the client's IP address needs to be sent to Okta using X-Forwarded-For.
  3.  When the application calls the Okta SDK, it explicitly passes the client IP from the incoming request to the SDK method.

Is that the use case you're solving for?

Okta.Core/Clients/AuthClient.cs
```diff
@@ -40,6 +40,49 @@ public virtual AuthResponse Authenticate(string username, string password, strin
...
+        /// <param name="password">User's password</param>
+        /// <param name="deviceToken">User's device token</param>
+        /// <param name="ipAddress">User's IP address</param>
+        /// <param name="userAgent">User's userAgent</param>
+        /// <param name="relayState">opaque value for the transaction and processed as untrusted data which is just echoed in a response. It is the client’s responsibility to escape/encode this value before displaying in a UI such as a HTML document </param>
+        /// <param name="bWarnPasswordExpired">Optional parameter indicating whether the PASSWORD_WARN status should be returned if available. Defaults to false</param>
+        /// <param name="bMultiOptionalFactorEnroll">Optional parameter indicating whether the user should be prompted to add an additional second factor if available </param>
+        /// <returns></returns>
+        public virtual AuthResponse AuthenticateWithContext(string username, string password, string deviceToken, string ipAddress = null, string userAgent = null, string relayState = null, bool bWarnPasswordExpired = false, bool bMultiOptionalFactorEnroll = false)
```
nbarbettini: Why don't we add a `AuthContext context` parameter to the existing `Authenticate` method, and push the concern of populating the AuthContext object up to the caller?

Okta.Core/OktaHttpClient.cs
```diff
@@ -175,6 +176,24 @@ public override Task<HttpResponseMessage> ExecuteAsync(HttpRequestType requestTy
...
+                            if (content.StartsWith("{") && content.EndsWith("}"))
+                            {
+                                var obj = JObject.Parse(content);
+                                ipAddress = (string)obj["context"]["ipAddress"];
+                            }
+                        }
+
+                        if (ipAddress != null)
+                        {
+                            httpClient.DefaultRequestHeaders.Add("X-Forwarded-For", ipAddress);
```
nbarbettini: Assuming the developer reuses the `IOktaHttpClient` object, this will add the header value to **all** subsequent requests. I think it should only be added to the current request. To do that, we'd have to construct an `HttpRequestMessage`.

Okta.Core/OktaHttpClient.cs
```diff
@@ -175,6 +176,24 @@ public override Task<HttpResponseMessage> ExecuteAsync(HttpRequestType requestTy
...
+                    if (bAddXForwardedHeader)
+                    {
+                        string ipAddress = null;
+                        if (!string.IsNullOrEmpty(content))
+                        {
+                            content = content.Trim();
+                            if (content.StartsWith("{") && content.EndsWith("}"))
+                            {
+                                var obj = JObject.Parse(content);
+                                ipAddress = (string)obj["context"]["ipAddress"];
```
nbarbettini: Deserializing our own content feels awkward. Ideally, we should refactor the `Post` method to accept an AuthContext, or some other entity that would contain this information. There may not be any easy way around this in the current architecture; thoughts @lboyette-okta or @jmelberg-okta?

Okta.Core/IOktaHttpClient.cs
```diff
@@ -18,10 +18,10 @@ public enum HttpRequestType
         }
 
         public virtual string ApiToken { get; set; }
-        public virtual Uri BaseUri { get; set; }
-
-        public abstract HttpResponseMessage Execute(HttpRequestType requestType, Uri uri = null, string relativeUri = null, string content = null, int waitMillis = 0, int retryCount = 0, bool bAddAuthorizationHeader = true);
-        public abstract Task<HttpResponseMessage> ExecuteAsync(HttpRequestType requestType, Uri uri = null, string relativeUri = null, string content = null, int waitMillis = 0, bool bAddAuthorizationHeader = true);
+        public virtual Uri BaseUri { get; set; }
+
+        public abstract HttpResponseMessage Execute(HttpRequestType requestType, Uri uri = null, string relativeUri = null, string content = null, int waitMillis = 0, int retryCount = 0, bool bAddAuthorizationHeader = true, bool bAddXForwardedHeader = false);
```
jmelberg-okta: @nbarbettini Since we are passing multiple headers, what about passing a dictionary instead?

Okta.Core/OktaHttpClient.cs
jmelberg-okta: Agreed. We shouldn't be deserializing the content here.

**Quick Fix**: At the very least, we should be able to `try catch` the `JObject.Parse` method, instead of checking for `{` and `}`.

Okta.Core/IOktaHttpClient.cs
nbarbettini: @jmelberg-okta The structure of how the SDK constructs requests should definitely be refactored, but I think that's outside the scope of this PR.

Okta.Core/Clients/AuthClient.cs
jyotsnaraghunathan-okta: Adding the AuthContext object as part of AuthnRequest doesn't really add the ip address as XFF header.  I tried doing that initially.

nbarbettini: You would still need the code you added to the `OktaHttpClient` class. This would just remove the need to add a separate `AuthenticateWithContext` method.

jyotsnaraghunathan-okta: Okay.  So instead of a new method in AuthClient.cs , just overload the existing Authenticate() ?

nbarbettini: @jyotsnaraghunathan-okta Yeah, with an `AuthContext` parameter.

jyotsnaraghunathan-okta: got it. Will make the change.

Okta.Core/OktaHttpClient.cs
jyotsnaraghunathan-okta: So what should I be doing for this ? Add a try catch around the JObject.Parse ?

Okta.Core/OktaHttpClient.cs
jyotsnaraghunathan-okta: Can you please provide more detail on what needs to be done here ?

Okta.Core/OktaHttpClient.cs
jmelberg-okta: I think that would be a bit cleaner than checking for brackets. Something along the lines of:
```csharp
try {
  var parsedContent = JObject.Parse(content);
  ipAddress = (string)parsedContent["context"]["ipAddress"];
  ...
} catch { /* no ipAddress given */}
```

Okta.Core/Clients/AuthClient.cs
nbarbettini: Since we made this change, can we remove the `AuthenticateWithContext` method now?

Okta.Core/Clients/AuthClient.cs
```diff
@@ -14,32 +15,93 @@ public class AuthClient : AuthenticatedClient
...
         /// <param name="bMultiOptionalFactorEnroll">Optional parameter indicating whether the user should be prompted to add an additional second factor if available </param>
-        /// <returns></returns>
-        public virtual AuthResponse Authenticate(string username, string password, string relayState = null, bool bWarnPasswordExpired = false, bool bMultiOptionalFactorEnroll = false)
-        {
-            var authRequest = new AuthRequest {
-                Username = username, 
-                Password = password, 
-                RelayState = relayState,
+        /// <returns></returns>
+        public virtual AuthResponse Authenticate(string username, string password, AuthContext authContext = null, string relayState = null, bool bWarnPasswordExpired = false, bool bMultiOptionalFactorEnroll = false)
```
nbarbettini: I think we should add the new `AuthContext` parameter at the end of the parameter list, so it's not a breaking change for existing calls.

Okta.Core/OktaHttpClient.cs
```diff
@@ -175,6 +176,29 @@ public override Task<HttpResponseMessage> ExecuteAsync(HttpRequestType requestTy
...
+                            content = content.Trim();
+                            try
+                            {
+                                var parsedContent = JObject.Parse(content);
+                                ipAddress = (string)parsedContent["context"]["ipAddress"];
+                            }
+                            catch (Exception e)
+                            {
+                                /* no ipAddress given */
+                                throw new OktaException("The AuthnRequest object is malformed." + e.Message);
```
nbarbettini: Is there ever a situation where the OktaHttpClient will POST some content that isn't JSON? (Maybe `application/x-www-form-urlencoded`? ping @jmelberg-okta )
If so, it will always throw an error here now.

Okta.Core/OktaHttpClient.cs
nbarbettini: Sure - we should not set `DefaultRequestHeaders`. Instead, we need to create a `new HttpRequestMessage` and set the header on that instance only. I can give you an example if you aren't familiar with the message object.

nbarbettini:
@jyotsnaraghunathan-okta Is this still waiting on a review from us?

nbarbettini:
.NET SDK versions 1.0+ support passing X-Forwarded-For and other headers along to Okta. I'm going to close this PR since there doesn't seem to be a strong need for it in older versions.

PR #58, opened 2017-06-13
Full rewrite (1.0)

nbarbettini:
Switching repo over to new code (1.x).


nbarbettini:
@lboyette-okta This has been squashed from 87 to 4 commits.

src/Okta.Sdk.Abstractions/HttpRequest.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public class HttpRequest
+    {
+        public string Uri { get; set; }
+
+        public object Payload { get; set; }
+
+        public IEnumerable<KeyValuePair<string, object>> QueryParams { get; set; }
+            = Enumerable.Empty<KeyValuePair<string, object>>();
+
```
jmelberg-okta: Should we add headers here?

src/Okta.Sdk/CastingListAdapter{T}.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+    public sealed class CastingListAdapter<T> : IList<T>
+    {
+        private readonly IList<object> _genericList;
+        private readonly ILogger _logger;
+
+        public CastingListAdapter(IList<object> genericList, ILogger logger)
+        {
+            _genericList = genericList;
+            _logger = logger;
```
jmelberg-okta: Why are these variables prepended with an underscore?
The [style guides](https://github.com/raywenderlich/c-sharp-style-guide#fields) I've seen recently highly discourage the use of underscores anywhere.

.gitattributes
```diff
@@ -0,0 +1,63 @@
...
+* text=auto
+
+###############################################################################
+# Set default behavior for command prompt diff.
+#
+# This is need for earlier builds of msysgit that does not have it on by
+# default for csharp files.
+# Note: This is only used by command line
+###############################################################################
+#*.cs     diff=csharp
```
jmelberg-okta: Shouldn't we remove all the commented out lines? Doesn't look like this file is doing us good at this point.

package.json
```diff
@@ -0,0 +1,24 @@
...
+  "main": "index.js",
+  "private": true,
+  "scripts": {
+    "build": "okta-sdk-generator -o sample -t templates"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/okta/oktasdk-csharp.git"
+  },
+  "author": "Len Boyette <boyettel@gmail.com> (http://lenboyette.com)",
```
jmelberg-okta: Lets remove Len's info here.

src/Okta.Sdk.Abstractions/HttpResponse{T}.cs
```diff
@@ -0,0 +1,25 @@
+﻿// <copyright file="HttpResponse{T}.cs" company="Okta, Inc">
+// Copyright (c) 2014-2017 Okta, Inc. All rights reserved.
+// Licensed under the Apache 2.0 license. See the LICENSE file in the project root for full license information.
+// </copyright>
+
+using System.Collections.Generic;
+
+namespace Okta.Sdk.Abstractions
+{
+#pragma warning disable SA1402 // File may only contain a single type
```
jmelberg-okta: Is it common for `#pragma` to live outside of indentation?

.gitattributes
nbarbettini: I didn't create this file; I think it was generated by Visual Studio? I usually just leave it alone.

package.json
nbarbettini: 👍

nbarbettini:
Thanks for your review! There won't be any more PRs quite as massive as this one.

> Does the {T}.cs indicate that it was a generated file?

No, that's a generic class. There are three main approaches for naming a class like `MyClass<T>`:
- `MyClass.cs` - not obvious from the file name that the class is generic
- `MyClass`1.cs` - used internally in .NET sometimes
- `MyClass{T}.cs` - my preferred approach, and the one I've used here

nbarbettini:
> Can we put a version number someone (package.json would work) to indicate that it is an alpha release?

Yep, added. 👍 There will be more package info added before we actually push this to NuGet.

src/Okta.Sdk.Abstractions/HttpRequest.cs
nbarbettini: These are just the empty defaults for a generic request. The RequestExecutor automatically adds a few default headers such as User-Agent.

src/Okta.Sdk/CastingListAdapter{T}.cs
nbarbettini: I'm following the [style](https://github.com/aspnet/Home/wiki/Engineering-guidelines#coding-guidelines) that Microsoft uses for ASP.NET Core: `_camelCase` for private fields.

PR #99, opened 2017-06-30
small changes to integration tests

leebrandt:


src/Okta.Sdk.IntegrationTests/UserScenarios.cs
```diff
@@ -36,16 +36,22 @@ public class UserScenarios : ScenarioGroup
...
-                .ToArray();
-
-            foundUsers.Length.Should().Be(1);
-            foundUsers.Single().Id.Should().Be(user.Id);
-
-            await user.DeactivateAsync();
-            await user.DeactivateOrDeleteAsync();
+            try
+            {
+                await Task.Delay(1000);
```
nbarbettini: Can you add a comment here to explain why this delay is necessary?

src/Okta.Sdk.IntegrationTests/UserScenarios.cs
```diff
@@ -36,16 +36,22 @@ public class UserScenarios : ScenarioGroup
...
+                var foundUsers = await client
+                    .Users
+                    .ListUsers(search: $"profile.nickName eq \"{user.Profile.GetProperty<string>("nickName")}\"")
+                    .ToArray();
+
+                foundUsers.Length.Should().Be(1);
+                foundUsers.Single().Id.Should().Be(user.Id);
+            }
+            finally
```
nbarbettini: 👍 I like this try/catch/finally - can you add this to all of the tests in this file? This will prevent any issues where tests fail and then aren't cleaned up.

PR #105, opened 2017-07-05
More IT scenarios

nbarbettini:
Adds a number of test scenarios, as well as helper methods/classes to make the SDK API better for those scenarios.


nbarbettini:
TODO: make sure this doesn't overwrite any tests Lee just merged.

PR #141, opened 2017-07-17
Preserve existing files when generating autodocs

### LINKED ISSUE #134: Add redirect from root to /latest
If you navigate to the root of the autodocs, you'll see a 404. We should add a redirect to `/latest`.

### LINKED ISSUE #139: Preserve old doc versions when new autodocs are generated

nbarbettini:
Improvements to the autodoc publishing process.

Closes #134 
Closes #139

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,5 +1,26 @@
+#addin "Cake.Git"
+#addin "Cake.GitPackager"
 #addin "Cake.DocFx"
+#addin "Cake.FileHelpers";
 #tool "docfx.console"
+
+
+// Helper method for setting a lot of file attributes at once
+public FilePath[] SetFileAttributes(FilePathCollection files, System.IO.FileAttributes fileAttributes)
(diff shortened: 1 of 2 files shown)


build.cake
```diff
@@ -69,22 +90,46 @@ Task("BuildDocs")
...
+        DeleteDirectory(tempDir, recursive: true);
     }
-    
-    EnsureDirectoryExists("./docs/temp");
+
+    GitClone("https://github.com/okta/okta-sdk-dotnet.git",
+            tempDir,
+            new GitCloneSettings
+            {
+                BranchName = "gh-pages",
```
jmelberg-okta: What happens if this branch does not exist?

nbarbettini: The build would definitely break 😄 this shouldn't happen, since I already created the `gh-pages` branch. If that gets deleted, we have bigger problems.

PR #144, opened 2017-07-17
Don't use partial updates for User model

### LINKED ISSUE #143: UpdateUserAsync throws an error
The `UserScenarios.UpdateUserProfile` IT is failing because the SDK is sending a PUT (full update) but the User object is internally tracked with changes (partial update).

nbarbettini:
Small change to the template so that the User model doesn't expect a partial update.

Closes #143


nbarbettini:
Thanks! It does now 😉

PR #160, opened 2017-10-31
Pin cake addin versions

nbarbettini:
Fixed the broken travis builds by pinning known good versions of cake addins.


nbarbettini:
Two addins got auto-updated: Cake.Git, Cake.FileHelpers. Both versions included breaking changes :(

As a precaution, I pinned everything so it keeps working indefinitely.

PR #163, opened 2017-11-08
Factors API

lboyette-okta:


src/Okta.Sdk/Internal/ResourceTypeResolver.cs
```diff
@@ -0,0 +1,46 @@
...
+                }
+
+                var baseTypeInfo = typeInfo.BaseType.GetTypeInfo();
+                var inheritsFromAbstractResolver = baseTypeInfo.IsGenericType
+                    && baseTypeInfo.GetGenericTypeDefinition() == typeof(AbstractResourceTypeResolver<>);
+
+                return inheritsFromAbstractResolver;
+            })
+            .Select(typeInfo => (typeInfo.AsType(), typeInfo.BaseType.GenericTypeArguments[0]))
+            .ToArray();
```
jmelberg-okta: Is it possible for this chain to produce `null` at this point? Looks like when `typeInfo.BaseType == null`, we return `false`, which gets passed into a failing `.Select()` method.

src/Okta.Sdk.IntegrationTests/FactorScenarios.cs
```diff
@@ -0,0 +1,96 @@
...
+            var profile = new UserProfile
+            {
+                FirstName = "Jill",
+                LastName = "Factor-SecurityQuestion",
+                Email = "jill-factor-securityquestion@example.com",
+                Login = "jill-factor-securityquestion@example.com",
+            };
+            profile["nickName"] = "jill-factor-securityquestion";
+
+            var createdUser = await client.Users.CreateUserAsync(new CreateUserWithPasswordOptions
```
lboyette-okta: This needs to be in the `try`

lboyette-okta: Nevermind

src/Okta.Sdk/Internal/ResourceTypeResolver.cs
nbarbettini: Returning `false` from inside the `Where` will simply skip, so that item never makes it to `Select`.

`typeInfo` should never be null, but it could be guarded against by changing the if to `typeInfo?.BaseType == null`

package.json
```diff
@@ -15,7 +15,7 @@
   },
   "homepage": "https://github.com/okta/oktasdk-sdk-dotnet#readme",
   "devDependencies": {
-    "@okta/openapi": "0.6.0",
+    "@okta/openapi": "^0.8.0",
```
nbarbettini: My preference is to explicitly pin this so it never changes without us knowing. Is there a reason you want `^`?

lboyette-okta: nope, good catch

src/Okta.Sdk/Internal/ResourceTypeResolver.cs
lboyette-okta: added Nate's guard suggestion

PR #170, opened 2018-01-24
Address legacy vs latest SDK confusion

lboyette-okta:


lboyette-okta:
We may also want to add examples for how to use the Groups and Factors APIs so people don't think they're limited to the Users API

nbarbettini:
I'd suggest changing this

> The latest SDK is where all the new and exciting development will happen.

to 

> This is the latest version of the .NET SDK (1.x), and we'll be adding many new features soon.

Or something to that effect. It clarifies that 1.x is "latest".

PR #171, opened 2018-03-06
:bug: Fix Powershell module TLS 1.2 compatibility

laurarodriguez-okta:
Adding support for TLS 1.2.

Resolves: OKTA-161239


laurarodriguez-okta:
@nbarbettini

nbarbettini:
As an extra experiment, can you see if this works?

```csharp
ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
```

This will turn on TLS 1.2 support without excluding any future changes to the default. There's some background in this Stack Overflow answer: https://stackoverflow.com/a/36454717/3191599

nbarbettini:
We could also consider bumping the `TargetFrameworkVersion` to 4.5 or 4.6 instead of 4.0. However, that would require some more testing to make sure we don't mess existing use cases up. Setting `ServicePointManager.SecurityProtocol` seems like an easy fix for now.

laurarodriguez-okta:
Both orgs (TLS 1.2 only & !TLS 1.2 only) works fine after updating the code to turning on TLS 1.2 without affecting other protocols.

nbarbettini:
Awesome! A few minor tasks and then we're done:

- [ ] Bump the version number to `1.0.1`: https://github.com/okta/okta-sdk-dotnet/blob/legacy/Okta.Core.Automation/Okta.Core.Automation.psd1#L3
- [ ] Update the changelog: https://github.com/okta/okta-sdk-dotnet/blob/legacy/Okta.Core.Automation/changelog.md
- [ ] Publish the new version to PowerShell Gallery

laurarodriguez-okta:
I don't have permissions to merge this.

PR #174, opened 2018-04-10
Update readme with more sdk examples

laura-rodriguez:
* Add examples of groups management
* Add TOC

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,29 @@ The SDK is compatible with:
 * .NET Framework 4.6.1
 * Mono
 
-## :warning: :construction: Alpha Preview :construction: :warning:
+## Table of Contents
+
+<!-- vscode-markdown-toc -->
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -139,6 +139,8 @@ vader.Profile["nickName"] = "Lord Vader";
 var newVader = await vader.UpdateAsync();
```
laura-rodriguez: Change this to `await vader.UpdateAsync();`

laura-rodriguez: After this section, add a new section: **Get and set custom attributes**

nbarbettini: Still need to update line 117, and add a section before line 120:

```
### Get and set custom attributes
(some explanation)

vader.Profile["homeworld"] = "Tattooine";
```

The note on line 120 will follow this 👍

Issue #156, opened 2017-09-14
Throttling Controls


shoorik:
Per [Okta's rate-limiting recommendations](https://developer.okta.com/docs/api/getting_started/design_principles#rate-limiting), it would be nice if there was a mechanism for the caller to define behavior when Okta SDK nears or runs afoul of the limit.

Perhaps the caller could set a configuration option `OverRateLimitBehavior` (when the response status code is 429), with possible values of `PauseAndRetry` (default) or `Fail`. The `PauseAndRetry` option would sleep until `X-Rate-Limit-Reset` and repeat the request.

For cases where the caller is nearing the limit, perhaps they could set an option of `RateLimitRemainingMin` which would pause until `X-Rate-Limit-Reset` if the `X-Rate-Limit-Remaining` falls below the "floor" to prevent "Rate limit warning" from being logged to sys log.

At the very least, if the pause behavior is not implemented, perhaps the helper methods from different `*Client` classes could expose the `X-Rate-Limit-*` headers in some sort of a statistics class attached to result objects, so that callers could have a chance to rate-limit themselves, without punting to the low-level Get/Put/Post methods..

nbarbettini:
Thanks for your suggestions (and your patience) @shoorik!

I agree that the SDK should make this easy to handle. I think the first pass will be automatically retrying, and future passes can add configuration. I'll track changes in this issue.

mbegan:
@nbarbettini very interested in the approach taken, I have need to incorporate in other integrations.

For reference (no laughing please) this is what I did to back-off and avoid rate limit exhaustion
https://github.com/mbegan/Okta-PSModule/blob/master/Okta.psm1#L272

nbarbettini:
FYI - we are planning on working on this in the near future. Stay tuned here 👍

agiles231:
I have written a solution called proktasy that acts a proxy between Okta REST API clients and Okta. It is easily deployed using Tomcat and resolves all rate limiting issues. Also, an SDK would have a difficult time accounting for rate limiting produced by other applications, even if using the same SDK. proktasy solves that too.

laura-rodriguez:
Hi Everyone 😄 ! We already provide a way to handle rate limit scenarios. Check out the readme for more info: https://github.com/okta/okta-sdk-dotnet#rate-limiting.

PR #181, opened 2018-05-17
:seedling: Adds support for handling API-rate limiting

ryannewington:
Adds support for detecting rate-limit errors and retrying operations. The code detects reset times provided by the API, and uses an exponential back-off algorithm. The code distinguishes from org-wide rate limits, as well as concurrent operation limits, handling both appropriately. In order to implement the retry functionality, calls to SendAsync were modified to pass in a HttpRequestMessage factory, rather than a HttpRequestMessage itself, as the HttpRequestMessage can only be used once. The OktaClientConfiguration object can be used to specify the number of retry operations permitted (default is 8)

Resolves: #156

### DIFF
--- a/src/Okta.Sdk/Configuration/OktaClientConfiguration.cs
+++ b/src/Okta.Sdk/Configuration/OktaClientConfiguration.cs
@@ -54,6 +54,14 @@ namespace Okta.Sdk.Configuration
         /// <remarks>An API token can be generated from the Okta developer dashboard.</remarks>
         public string Token { get; set; }
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


nbarbettini:
This is great code, thanks! We will most likely use a version of it for official retry support in the SDK.

To give you some background, implementing retries (and some other logic) is on our roadmap but we're still spec'ing it out. We're going to add an interface to allow the developer to customize the retry behavior if they want. I'm guessing it'll be a couple of months based on other priorities.

I've copied your work into this branch: https://github.com/okta/okta-sdk-dotnet/tree/add-rate-limiting  so it doesn't get lost. I'll close the PR for now, but I'll leave the issue open. Let me know if you have any questions!

Issue #176, opened 2018-04-19
Expose errorCauses


HakanL:
I tried this package the other day and did some tests with creating a new user. First I got an generic E0000001 API Validation exception back, without any more details. It turns out I didn't set the Login property. Shouldn't there be more details/checks for a condition like that?

Second I tried a simple password (knowing it would fail the Okta policy) and I got an exception back with a bunch of formatted text with mixed text about policy, callstack, etc. Shouldn't an exception like this be in a way where it can be shown to the user, or interpreted in code? They need to know what password policy they broke so they can fix it.

nbarbettini:
I agree, these exceptions aren't ideal. Unfortunately the SDK is limited by the error information returned by the [API](https://developer.okta.com/reference/error_codes/), which is currently pretty terse.

The `OktaApiException` class exposes everything it can (StatusCode, ErrorCode, ErrorSummary, etc) but that's the best we can do right now.

I'd recommend posting a feature request over on the [Dev Forum](https://devforum.okta.com/c/feature-requests) for more detailed API error messages. 👍

nbarbettini:
Closing since there isn't anything the SDK can do about this (right now at least).

HakanL:
For reference: https://devforum.okta.com/t/better-exceptions/1642

nbarbettini:
Awesome, thanks @HakanL!

HakanL:
Some on the developer forum indicated that there is a property with the user-friendly error called errorCauses. See the dev forums link above. Can we re-open this ticket?

nbarbettini:
Aha, I forgot about this comment 😉 https://github.com/okta/okta-sdk-dotnet/blob/master/src/Okta.Sdk/OktaApiException.cs#L67

Exposing `errorCauses` makes sense.

PR #182, opened 2018-05-17
:seedling: Adds support for error causes in OktaApiException

ryannewington:
Exposes the errorCauses element of an API response in the OktaApiException, and includes this information in the exception's Message property to surface the information with standard logging libraries that call ToString() on an exception.

Resolves: #176


nbarbettini:
Hi again @ryannewington! Do you mind following [these instructions](https://developer.okta.com/cla/) to sign our CLA?

ryannewington:
@nbarbettini done

nbarbettini:
Merged via #185

PR #183, opened 2018-05-17
:seedling: Adds support for skinny_users endpoint

ryannewington:
Adds support for calling the skinny_users endpoint to get a smaller representation of the users who are a member of a group as documented in the following guide https://www.okta.com/sites/default/files/Okta_Security-Analytics_Partner-Integration-Guide.pdf

### DIFF
--- a/src/Okta.Sdk/Group.cs
+++ b/src/Okta.Sdk/Group.cs
@@ -16,6 +16,9 @@ namespace Okta.Sdk
         public IAsyncEnumerable<IUser> Users
             => GetClient().Groups.ListGroupUsers(Id);
 
+        public IAsyncEnumerable<IUser> UsersSkinny
+            => GetClient().Groups.ListGroupUsersSkinny(Id);
+
         /// <inheritdoc/>
         public Task DeleteAsync(CancellationToken cancellationToken = default(CancellationToken))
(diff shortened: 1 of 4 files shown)


nbarbettini:
I need to check w/ the team on adding this. If we decide to support the endpoint, we'll add it to the OpenAPI spec that creates the generated code files so that it is automatically supported in all SDKs.

In the meantime, would this be an acceptable workaround to having a `groups.UsersSkinny` property?

```csharp
var groupId = "00gblkmygroupid";
var skinnyUsers = client.GetCollection<IUser>($"/api/v1/groups/{groupId}/skinny_users");
```

ryannewington:
Yep, I'm more than happy to use that for now. 

What I really want is a `user_ids` endpoint. I've got groups with hundreds of thousands of users and I only need the user ID, not the whole user. It would save hundreds of megabytes of unnecessary data transfer, and not to mention be orders of magnitude faster. 

Is there a formal process for logging feature requests like this?

ryannewington:
Done. https://devforum.okta.com/t/new-endpoint-to-list-group-members-by-id-only/1850/1

Happy for you to reject this PR, the GetCollection method is working fine for me, and if it gets supported, I'll revert to using the generated method!

PR #184, opened 2018-05-17
:seedling: Adds support for disabling server certificate validation

ryannewington:
Adds support for disabling server certificate validation in the OktaClientConfiguration, which enables debugging scenarios using proxies such as fiddler.

### DIFF
--- a/src/Okta.Sdk/Configuration/OktaClientConfiguration.cs
+++ b/src/Okta.Sdk/Configuration/OktaClientConfiguration.cs
@@ -54,6 +54,14 @@ namespace Okta.Sdk.Configuration
         /// <remarks>An API token can be generated from the Okta developer dashboard.</remarks>
         public string Token { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether the server's TLS certificate is validated. Warning this should only be used for debugging purposes and never used in production.
+        /// </summary>
+        /// <value>
+        /// A value indicating whether the server's TLS certificate is validated
+        /// </value>
+        public bool DisableServerCertificateValidation { get; set; }
+
         /// <inheritdoc/>
         public OktaClientConfiguration DeepClone()
             => new OktaClientConfiguration
@@ -61,6 +69,7 @@ namespace Okta.Sdk.Configuration
                 ConnectionTimeout = ConnectionTimeout.HasValue ? this.ConnectionTimeout.Value : (int?)null,
                 OrgUrl = this.OrgUrl,
                 Proxy = this.Proxy?.DeepClone(),
+                DisableServerCertificateValidation = this.DisableServerCertificateValidation,
             };
     }
 }
--- a/src/Okta.Sdk/Internal/DefaultRequestExecutor.cs
+++ b/src/Okta.Sdk/Internal/DefaultRequestExecutor.cs
@@ -51,7 +51,8 @@ namespace Okta.Sdk.Internal
                 configuration.Token,
                 configuration.ConnectionTimeout,
                 configuration.Proxy,
-                logger);
+                logger,
+                configuration.DisableServerCertificateValidation);
         }
 
         private static string EnsureCorrectOrgUrl(string orgUrl)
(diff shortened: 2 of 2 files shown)


nbarbettini:
I understand the scenario where this is useful, but I'm wary of supporting a certificate workaround as "official" behavior. I'd rather make it possible to pass your own `HttpClientHandler` to the client if you want to customize this.

The only client code that would need to be refactored is here: https://github.com/okta/okta-sdk-dotnet/blob/master/src/Okta.Sdk/Internal/DefaultRequestExecutor.cs#L84-L92

What do you think?

ryannewington:
Let me try that. I spent a lot of time trying different ways to do this. It's actually supported to do this app-wide in .NET framework using this property
https://msdn.microsoft.com/en-us/library/system.net.servicepointmanager.servercertificatevalidationcallback(v=vs.110).aspx

But it doesn't pass into the .NET Standard world and is ignored. I encountered quite a few issues 'passing over the boundary' between the two worlds. Let me see what luck I can have with the HttpClient directly.

ryannewington:
Yeah that doesn't work it seems. I got this message a few times with different variants of trying to solve this problem. The other thing I tried was just exposing the callback in the OktaClientConfiguration, but that didn't work either.

```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.Net.Http.HttpClientHandler.get_CheckCertificateRevocationList()
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index)
   at FlexibleConfiguration.Providers.ObjectVisitors.ObjectVisitor.VisitProperties(Object obj)
   at FlexibleConfiguration.Providers.ObjectVisitors.ObjectVisitor.VisitObject(Object obj)
   at FlexibleConfiguration.Providers.ObjectVisitors.ContextAwareObjectVisitor.VisitProperty(String name, TypeInfo propertyTypeInfo, Object actualValue)
   at FlexibleConfiguration.Providers.ObjectVisitors.ObjectVisitor.VisitProperties(Object obj)
   at FlexibleConfiguration.Providers.ObjectVisitors.ContextAwareObjectVisitor.Visit(Object obj)
   at FlexibleConfiguration.Providers.ObjectReflectionConfigurationProvider.Load()
   at FlexibleConfiguration.ConfigurationBuilder.Add(IConfigurationProvider provider)
   at Okta.Sdk.OktaClient.CompileFromConfigurationSources(OktaClientConfiguration apiClientConfiguration)
   at Okta.Sdk.OktaClient..ctor(OktaClientConfiguration apiClientConfiguration, ILogger logger)
   at Lithnet.Okta.ManagementAgent.Ecma2.OpenConnection(KeyedCollection`2 configParameters) in D:\github\lithnet\okta-managementagent\Lithnet.Okta.ManagementAgent\Ecma2.cs:line 499
   at Lithnet.Okta.ManagementAgent.Ecma2.GetSchema(KeyedCollection`2 configParameters) in D:\github\lithnet\okta-managementagent\Lithnet.Okta.ManagementAgent\Ecma2.cs:line 333
```

nbarbettini:
I think with the new [HttpClientFactory pattern](https://blogs.msdn.microsoft.com/webdev/2018/02/28/asp-net-core-2-1-preview1-introducing-httpclient-factory/) coming down the pipe, it makes sense that the `HttpClient` (which includes the `HttpClientHandler`) the right extensibility point. 

I think it makes sense to:
* Add the ability for a developer to optionally pass in their own `HttpClient` to be used by the SDK
* Fall back to a default client if no client is provided (much like today)

nbarbettini:
Closing this in favor of #190 (which builds on this PR). Thanks for all the contributions @ryannewington!

Issue #187, opened 2018-05-22
Creating a user with provider options does not set url parameter of provider=true


ryannewington:
The CreateUserAsync overload that takes a CreateUserWithProviderOptions does not pass through the optional provider=true parameter to the underlying CreateUserAsync method

nbarbettini:
Closed via #188

PR #188, opened 2018-05-22
:bug: Fixes an issue with creating users with a specific provider

ryannewington:
Fixes issue #187 where the CreateUserAsync overload that takes a CreateUserWithProviderOptions does not pass through the optional provider=true parameter to the underlying CreateUserAsync method

### DIFF
--- a/src/Okta.Sdk/UsersClient.cs
+++ b/src/Okta.Sdk/UsersClient.cs
@@ -109,7 +109,7 @@ namespace Okta.Sdk
                 },
             };
 
-            return CreateUserAsync(user, options.Activate, cancellationToken: cancellationToken);
+            return CreateUserAsync(user, options.Activate, true, cancellationToken);
         }
 
         /// <inheritdoc/>


nbarbettini:
Thanks for catching this! You'll need to update this branch (the big button above), then feel free to merge. If git doesn't cooperate, let me know.

ryannewington:
I could update the branch, but I'm not allowed to merge

nbarbettini:
Done!

ryannewington:
Awesome. PS I only found this bug because I had #184 enabled and saw it in fiddler :)

nbarbettini:
@ryannewington Are you able to use Fiddler's HTTPS decryption option? Or do you not want to trust their root cert?

ryannewington:
It's a matter of trusting their root cert system-wide versus just for a single application. The MIM system usually has high level access to lots of different systems for provisioning purposes, and security is a concern. For helping customers troubleshoot, being able to set a 'debug' flag and have them send me outputs from fiddler traces is very valuable.

Alternatively, is there a way to log raw request/response data with the SDK? This would avoid having to do any certificate trickery at all.

PR #190, opened 2018-06-06
Allow developer to pass in HttpClient

nbarbettini:
Adds an overload to allow a developer to pass in their own `HttpClient`. This is useful for:

* Adding custom logging or request behavior
* Use with the [HttpClientFactory pattern](https://blogs.msdn.microsoft.com/webdev/2018/02/28/asp-net-core-2-1-preview1-introducing-httpclient-factory/) in ASP.NET Core 2.1
* Customizing the `HttpClientHandler`, like in https://github.com/okta/okta-sdk-dotnet/pull/184

Tested by
- [x] Running integration tests
- [x] Manually trying an HTTP client with custom behavior


nbarbettini:
@ryannewington Based on your request in #184 and our discussion there, I came up with this. If you need to disable certificate validation, you could:

```csharp
// Set up an HttpClient with custom behavior
var handler = new HttpClientHandler
{
    AllowAutoRedirect = false,
    ServerCertificateCustomValidationCallback = (httpRequestMessage, x509Certificate2, x509Chain, sslPolicyErrors) => true
};
var httpClient = new HttpClient(handler, true);

// Pass it to OktaClient
var oktaClient = new OktaClient(oktaConfig, httpClient);
```

This way, disabling validation isn't baked directly in the SDK, but you can do it if you want (and understand the risks). Does this direction make sense for your use case?

ryannewington:
@nbarbettini this is definitely a worthwhile feature, and should be included, but it's not suitable for me because trying to access the .Net Core HttpClient in .NET framework throws an exception.

System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.PlatformNotSupportedException: Operation is not supported on this platform.

ryannewington:
I'm happy to keep my own fork of the SDK to do what I need to do though. 

We probably just need to check that bubbling up the HttpClient doesn't make the SDK inaccessible to .net fx clients. Happy to do some testing on this.

nbarbettini:
Curious, which property throws that `PlatformNotSupportedException`? I'm testing it with .NET Framework 4.6.1 and so far it seems to be working. I'm sure I'm not hitting some .NET Core-only property.

Since [System.Net.Http](https://www.nuget.org/packages/System.Net.Http) targets .NET Framework 4.5, 4.6, and .NET Standard 1.1+, bubbling up the `HttpClient` to the constructor should be safe. If you could test it yourself, that would be great!

nbarbettini:
Ah, that makes sense. The configuration code does some weird reflection-y things when you add properties to configuration. It must have been doing something that `HttpClient` didn't like. The way I've implemented it, it shouldn't be a problem for normal use cases on both .NET Framework and .NET Core.



> But it probably just needs to be called out in the doco somewhere that certain properties of the OktaClientConfiguration are ignored if using a custom HttpClient.

Agreed!

Issue #201, opened 2018-06-13
Add Apps API


nbarbettini:
Apps management API: https://developer.okta.com/docs/api/resources/apps

- [x] Update generated code to spec 0.9.0
- [x] Clean up code issues
- [x] Write integration tests (#200)
- [x] Need to add a resolver?
- [x] Define what helpers should be added
- [ ] Update generated code to spec 1.1.0 (after https://github.com/okta/okta-sdk-dotnet/issues/203)

PR #208, opened 2018-06-21
:seedling: Exposes the next link header as a continuation token

dnlhll:
Exposes the next link header (https://developer.okta.com/docs/api/getting_started/design_principles#pagination) as a continuation token when listing users or groups

### DIFF
--- a/src/Okta.Sdk/GroupsClient.cs
+++ b/src/Okta.Sdk/GroupsClient.cs
@@ -7,6 +7,10 @@ using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 
+using Newtonsoft.Json;
+
+using Okta.Sdk.Internal;
+
 namespace Okta.Sdk
 {
     /// <summary>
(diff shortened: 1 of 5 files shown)


nbarbettini:
Hey @dnlhll, thanks for your contribution! ❤️ I'll do a review of the code.

nbarbettini:
Can you give an example of how you're using this? What does the SDK not do today that you needed to solve?

nbarbettini:
Got it, thanks for the detailed explanation! Mostly I wanted to make sure you were aware that `await client.Users.ToList/ToArray()` worked already (since it's not well-documented yet). 🙂 

To make sure I am understanding your two use cases correctly:
* In scenario #1, `ToArray()` won't work because you explicitly want to retrieve one page at a time and hand the downstream caller a `next` cursor.
* In scenario #2, you can use `ToArray()` (the async enumerator) but you want to be able to grab the cursor as pages are enumerated so you can pick up where you left off if something goes wrong.

For #2, I want to investigate using the [IProgress<T>](https://blogs.msdn.microsoft.com/dotnet/2012/06/06/async-in-4-5-enabling-progress-and-cancellation-in-async-apis/) approach. It's already built-in and I think it would make it easy to "subscribe" to the cursor if you need to.

dnlhll:
Yes, I think you've got the gist of it.

Scenario 1: You are right, using `.ToList()/.ToArray()` might cause too much memory pressure by pulling down the entire entity set. I'd prefer to work with smaller batches.

Scenario 2: For the second scenario, passing an `IProgress` arg makes enough sense assuming the `Report(T)` method would be passed the cursor on each new page that is pulled down by the `IAsyncEnumerator`. However, if my process crashes, I still need a method where I can pass the cursor as a starting point.

The simplest explanation is that the Okta REST API gives consumers control of pagination and the C# client does not. That feature can be exposed directly (sort of my approach by just returning a poco with the exact contents of the REST API response) or it can be translated into a more OO C# experience using .NET interfaces like `IProgress` and `IAsyncEnumerator`. Returning the exactly contents of the REST API response means features wont be accidentally hidden, but the experience is less idiomatic than wrapping with .NET interfaces. You have more context on how people are using this client, so i'll leave the design decisions to you.

dnlhll:
@nbarbettini, do you have any thoughts or questions about my most recent clarification? Do the use cases make sense?

nbarbettini:
@dnlhll Your clarification makes sense! Sorry for the radio silence. We've been focused on adding the Applications API (#201) and prepping for a new release of the SDK. I don't think this addition will make it into the next incremental (alpha8) release, but I do want to add this soon.

In the meantime, are you still able to use a private build so you are at least not blocked?

dnlhll:
@nbarbettini, no worries. The private build will do for now, as long as the feature is being tracked in some way.

nbarbettini:
Superseded by https://github.com/okta/okta-sdk-dotnet/pull/253

@dnlhll Whenever you get a chance, let me know if the additions in #253 cover your needs! If not, happy to take another look.

PR #219, opened 2018-07-09
Openapi 1.1.0

laura-rodriguez:
Bump spec version to 1.1.0

PR #223, opened 2018-07-12
Changes the `processModel` to generate `IList<Ix>` instead of `IList<x>`,

laura-rodriguez:


src/Okta.Sdk/Internal/CastingListAdapter{T}.cs
```diff
@@ -54,6 +54,8 @@ private T Cast(object item)
             // Fall back to primitive conversion
             try
             {
+                if (!(item is IConvertible)) { return (T)item; }
```
laura-rodriguez: Disregard this line.

nbarbettini: Is this part of another PR?

laura-rodriguez: No, this was an attempt to handle the exception `the object doesn't implement the `IConvertible` interface`

nbarbettini:
The problem was actually here: https://github.com/okta/okta-sdk-dotnet/pull/223/files#diff-8b7bc85bdb7632dc527405e6efdf2303L84

The factory didn't know how to "create" an interface. I added a lookup that will find the implementation `SamlAttributeStatement` given the interface `ISamlAttributeAttributeStatement`.

laura-rodriguez:
Merged in #219

PR #234, opened 2018-07-31
Add integration tests for group user operations

### LINKED ISSUE #93: Group integration test: Group User Operations
- [ ]  Create a user with credentials → POST /api/v1/users?activate=false

```js
const newUser = {
    profile: {
        firstName: 'John',
        lastName: 'With-Group',
        email: 'john-with-group@example.com',
        login: 'john-with-group@example.com'
    },
    credentials: {
        password: { value: 'Abcd1234' }
    }
};
```

- [x]  Create a new group → POST /api/v1/groups

```js
const newGroup = {
    profile: {
        name: 'Group-Member API Test Group'
    }
};
```

- [x] Add user to the group and validate user present in group → POST /api/v1/groups/{{groupId}}/users/{{userId}} + GET /api/v1/groups/{{groupId}}/users

- [x]  Remove user from group and validate user removed → DELETE  /api/v1/groups/{{groupId}}/users/{{userId}} + GET /api/v1/groups/{{groupId}}/users

- [x]  Delete the group and user → POST /api/v1/users/{{userId}}/lifecycle/deactivate + DELETE /api/v1/users/{{userId}} +  DELETE /api/v1/groups/{{groupId}}

laura-rodriguez:
Resolves
#93

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/GroupScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/GroupScenarios.cs
@@ -119,10 +119,224 @@ namespace Okta.Sdk.IntegrationTests
                 () => client.Groups.GetGroupAsync(createdGroup.Id));
         }
 
-        [Fact(Skip = "TODO")]
-        public async Task GroupUserOperations()
+        [Fact]
+        public async Task AddUserToGroup()
         {
-            throw new NotImplementedException();
+            var client = GetClient();
+
+            // Create group
+            var createdGroup = await client.Groups.CreateGroupAsync(new CreateGroupOptions
+            {
+                Name = "New Users Test Group",
+            });
+
+            // Create a user
+            var createdUser = await client.Users.CreateUserAsync(new CreateUserWithPasswordOptions
+            {
+                Profile = new UserProfile
+                {
+                    FirstName = "John",
+                    LastName = "Get-User",
+                    Email = "john-get-user@example.com",
+                    Login = "john-get-user@example.com",
+                },
+                Password = "Abcd1234",
+                Activate = true,
+            });
+
+            await client.Groups.AddUserToGroupAsync(createdGroup.Id, createdUser.Id);
+
+            try
+            {
+                var retrievedGroup = await client.Groups.GetGroupAsync(createdGroup.Id);
+                retrievedGroup.Should().NotBeNull();
+                var groupUsersList = await retrievedGroup.Users.ToList();
+                groupUsersList.Should().HaveCount(1);
+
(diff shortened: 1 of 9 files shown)


src/Okta.Sdk.IntegrationTests/GroupScenarios.cs
```diff
@@ -119,10 +119,224 @@ public class GroupScenarios : ScenarioGroup
...
+            });
+
+            // Create a user
+            var createdUser = await client.Users.CreateUserAsync(new CreateUserWithPasswordOptions
+            {
+                Profile = new UserProfile
+                {
+                    FirstName = "John",
+                    LastName = "Get-User",
+                    Email = "john-get-user@example.com",
```
nbarbettini: Rename email and login to `john-add-group@example.com` (these different emails are intended to reduce collisions if the IT doesn't clean up after itself properly for some reason)

src/Okta.Sdk.IntegrationTests/GroupScenarios.cs
```diff
@@ -119,10 +119,224 @@ public class GroupScenarios : ScenarioGroup
...
```
nbarbettini: Rename this email and login as well

src/Okta.Sdk.IntegrationTests/GroupScenarios.cs
```diff
@@ -119,10 +119,224 @@ public class GroupScenarios : ScenarioGroup
...
+                    Login = "john-get-user@example.com",
+                },
+                Password = "Abcd1234",
+                Activate = true,
+            });
+
+            await client.Groups.AddUserToGroupAsync(createdGroup.Id, createdUser.Id);
```
nbarbettini: I think these lines should also be within the `try`. That way if it fails at the beginning, things are still cleaned up. What do you think? (applies to all these new tests)

nbarbettini:
Looking good, a few comments. Can you split this up from the other code changes?

src/Okta.Sdk.IntegrationTests/GroupScenarios.cs
laura-rodriguez: if I want `createdUser` to be available in the `finally` scope, it should be declared before `try` block
@nbarbettini

nbarbettini: I should have been more specific - I meant just the `AddUserToGroup` line. You understood what I meant 🙂  👍

nbarbettini:
Superseded by #237

PR #240, opened 2018-08-03
Add implicit operator in the enum template.

### LINKED ISSUE #238: Implicit cast from string to StringEnum
It should be possible to do:

```
UserStatus foo = "ACTIVE";
foo == UserStatus.Active;
```

Since `StringEnum` is abstract, this will require a `implicit operator Xyz(string value)` in each file generated from `Enum.cs.hbs`. Should be straightforward to add to the template.

You can base your work on: https://github.com/okta/okta-sdk-dotnet/compare/nb-spike-stringenum-cast?expand=1

laura-rodriguez:
Resolves
#238

### DIFF
--- a/openapi/templates/Enum.cs.hbs
+++ b/openapi/templates/Enum.cs.hbs
@@ -18,6 +18,12 @@ namespace Okta.Sdk
 
         {{/each}}
         /// <summary>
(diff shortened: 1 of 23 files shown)


nbarbettini:
FYI, I think if you put `Resolves #238` on one line, it will close that issue automatically when this PR is merged.

PR #246, opened 2018-08-15
GetArrayProperty should not return null

### LINKED ISSUE #165: GetArrayProperty shouldn't return null
`GetArrayProperty` on resource types has a return type of `IList<T>`. Currently, if you request a resource property that doesn't exist, the method returns `null`. That's better than throwing an exception but still a little awkward for things like LINQ.

Some investigation is required, but I think the behavior should be: if the underlying property is null, return `new List<object>`

laura-rodriguez:
Resolves #165


src/Okta.Sdk.UnitTests/ApplicationsClientShould.cs
```diff
@@ -0,0 +1,103 @@
...
+            var mockRequestExecutor = new MockedStringRequestExecutor(GetBookmarkApplicationStubResponse());
+            var client = new TestableOktaClient(mockRequestExecutor);
+
+            var app = client.Applications.GetApplicationAsync<IBookmarkApplication>("foo").Result;
+
+            app.Features.Any().Should().BeFalse();
+        }
+
+        [Fact]
+        public void ReturnEmptyListWhenFeaturesIsNullInTheResponse()
```
nbarbettini: This test and the one above look the same to me. Should they be testing separate use cases?

laura-rodriguez: I forgot the '[]' in the getStub method.

PR #250, opened 2018-08-20
Add integration tests to update user profile with a dynamic object.

### LINKED ISSUE #152: Possible String URL Encoding Issue
oktaUser.Profile["custom-data"] = customData.ToString(Newtonsoft.Json.Formatting.None); 

await client.Users.UpdateUserAsync(oktaUser, idClaim.Value); 

With error: Okta.Sdk.OktaApiException: 'Api validation failed: userProfile'

laura-rodriguez:
Resolves #152


nbarbettini:
Looks good!

BTW, `var foo = new { ... }` is actually an **anonymous** type, not a dynamic type. There's some good info on this answer: https://stackoverflow.com/q/391529/3191599

PR #253, opened 2018-08-21
Manual pagination support

### LINKED ISSUE #222: Make headers or next link available during pagination
In most cases, this SDK handles [pagination](https://developer.okta.com/docs/api/getting_started/design_principles#pagination) of collections like Users and Groups for you automatically:

    // Get all users:
    var users = await oktaClient.Users.ToArray();

    // Or, do something on each user:
    await oktaClient.Users.ForEachAsync(user =>
    {
        // todo
    });

In these cases, the SDK is handling the enumeration/pagination for you transparently. This covers most things you'd need to do with Okta collections.

In less common cases, you might need to handle the pagination yourself at a lower level, by:

* Reading the link headers returned with a page of results to get a cursor
* Paging  (or resuming paging) the collection yourself

This SDK doesn't currently expose the [Link header](https://developer.okta.com/docs/api/getting_started/design_principles#link-header) returned by the Okta API. It uses it internally, but doesn't make it available to your code.

nbarbettini:
Resolves #222 

- [x] Split up existing internal enumerator to separate LINQ support (internal) and manual pagination (available to consumers)
- [x] Integration test for collection enumeration using LINQ
- [x] Integration test for manual collection enumeration by retrieving pages and resuming enumeration on a specific page

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/CollectionScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/CollectionScenarios.cs
@@ -26,6 +26,7 @@ namespace Okta.Sdk.IntegrationTests
         [Fact]
         public async Task EnumerateCollectionWithLinq()
         {
+            var successful = true;
             var client = TestClient.Create();
             var createdUsers = new ConcurrentBag<IUser>();
 
@@ -48,6 +49,7 @@ namespace Okta.Sdk.IntegrationTests
             catch (Exception e)
             {
                 _logger.LogMessage($"Exception while running test: {e.Message}");
+                successful = false;
             }
             finally
             {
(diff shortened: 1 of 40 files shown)


nbarbettini:
Yeah, pushing a few more commits soon to finish this PR.

nbarbettini:
Ready for review!

src/Okta.Sdk.IntegrationTests/CollectionScenarios.cs
```diff
@@ -56,6 +58,121 @@ public async Task EnumerateCollectionWithLinq()
...
+
+                while (await enumerator.MoveNextAsync())
+                {
+                    retrievedUsers.AddRange(enumerator.CurrentPage.Items);
+                }
+
+                retrievedUsers.Count.Should().Be(10);
+            }
+            catch (Exception e)
+            {
```
laura-rodriguez: Don't you have the exception when it fails already?

dnlhll:
@nbarbettini does this cover the feature suggested in this pr? https://github.com/okta/okta-sdk-dotnet/pull/208

Am I understanding correctly that the test ‘EnumerateCollectionManuallyAndResume’ is showing this functionality?

src/Okta.Sdk.IntegrationTests/CollectionScenarios.cs
nbarbettini: Without this `catch`, the test will appear to pass even if an error occurs, because the `finally` block will swallow the exception and complete successfully. This `catch` and `successful.Should().BeTrue()` at the end ensures the test shows up as a failure if something goes wrong.

nbarbettini:
@dnlhll Yep! Sorry again for the delay. I took some of your PR plus some internal refactoring to make your use cases possible. The [new integration tests](https://github.com/okta/okta-sdk-dotnet/pull/253/files#diff-1046a917b648fba44b2c041114136bf2) show how the syntax will work.

Does this look like it covers your needs?

nbarbettini:
@dnlhll Great! Post more issues for anything else you run into or need.

PR #258, opened 2018-08-23
Runs integration tests on travis

vijetmahabaleshwar-okta:
* Adds support to run integration tests for PRs
* Will only run ITs for PRs and cron jobs, not for push requests
* If we run ITs for push requests as well, since both builds (push and pull) will be running simultaneously against the same trex org, tests will fail (with resource already exists error)


src/Okta.Sdk/Internal/OktaClientConfigurationValidator.cs
```diff
@@ -29,7 +29,7 @@ public static void Validate(OktaClientConfiguration configuration)
 
             if (!configuration.OktaDomain.StartsWith("https://", StringComparison.OrdinalIgnoreCase))
             {
-                throw new ArgumentNullException(nameof(configuration.OktaDomain), "Your Okta URL must start with https. You can copy your domain from the Okta Developer Console.");
+                throw new ArgumentException(nameof(configuration.OktaDomain), "Your Okta URL must start with https. You can copy your domain from the Okta Developer Console.");
```
vijetmahabaleshwar-okta: This change is needed to get the following IT to pass - `Okta.Sdk.IntegrationTests/OktaClientShould.cs`
It expects the `ArgumentException`.

src/Okta.Sdk.IntegrationTests/OktaClientShould.cs
```diff
@@ -71,6 +71,7 @@ public async Task ThrowApiExceptionForInvalidToken()
             var client = TestClient.Create(new OktaClientConfiguration
             {
                 Token = "abcd1234",
```
laura-rodriguez: In this test, we shouldn't enable the flag. Instead, we have to add a valid oktaDomain.

src/Okta.Sdk.IntegrationTests/LogsScenarios.cs
```diff
@@ -99,6 +99,13 @@ public async Task GetLogsByEventType()
         [Fact]
         public async Task GetLogsBySinceDate()
         {
+            // Don't run test on travis - https://github.com/okta/okta-sdk-dotnet/issues/271
```
vijetmahabaleshwar-okta: This test always fails on travis. I've created a github issue to track it.

nbarbettini:
@vijetmahabaleshwar-okta Let me know when you want me to re-review this. (And I agree with Laura, this branch should be updated/rebased to latest master)

laura-rodriguez:
We would be able to use `NewGuid()` in resources name to avoid:
_If we run ITs for push requests as well, since both builds (push and pull) will be running simultaneously against the same trex org, tests will fail (with resource already exists error)_
But is it the solution @nbarbettini? Other than that, it looks good to me 👍.

build.sh
```diff
@@ -22,9 +22,16 @@ else
     MD5_EXE="md5sum"
 fi
 
-# Define default arguments.
+# Define default arguments
 SCRIPT="build.cake"
-TARGET="Default"
+TARGET=
+# Don't run ITs for push requests
+if [[ $TRAVIS_EVENT_TYPE != 'push' ]]; then
```
nbarbettini: Why run ITs by default at all? You should be able to manually do:

```
./build.sh --target DefaultIT
```

src/Okta.Sdk.IntegrationTests/LogsScenarios.cs
```diff
@@ -103,6 +103,13 @@ public async Task GetLogsByEventType()
         [Trait("Category", "NoBacon")] // Tests that don't run on internal CI pipeline
         public async Task GetLogsBySinceDate()
         {
+            // Don't run test on travis - https://github.com/okta/okta-sdk-dotnet/issues/271
+            string travis = System.Environment.GetEnvironmentVariable("TRAVIS");
```
nbarbettini: Let's `[Skip]` tests that are currently broken rather than hard code env var checks.

vijetmahabaleshwar-okta: Done

nbarbettini:
@laura-rodriguez 
> We would be able to use NewGuid() in resources name to avoid:
> _If we run ITs for push requests as well, since both builds (push and pull) will be running simultaneously against the same trex org, tests will fail (with resource already exists error)_
I agree, I should have built them that way (with random GUIDs) from the beginning. Otherwise the IT suite is always a "singleton" with respect to the org. We should be able to run tests in parallel as much as we want (because the tests clean up properly _and_ use randomized emails). Separate PR?

PR #261, opened 2018-08-28
Use diff username in application tests

laura-rodriguez:
Use different user emails to avoid possible conflicts.


laura-rodriguez:
Replaced by #263

PR #273, opened 2018-09-13
Adds category for tests that don't run on CI

vijetmahabaleshwar-okta:


nbarbettini:
Is there a path to getting these tests running later? Or is it impossible?

vijetmahabaleshwar-okta:
> Is there a path to getting these tests running later? Or is it impossible?

Not sure if there's a way to get them running on bacon. Have to look into it. So far, it's not possible.
We can run these tests against trex/preview as nightly jobs though. So, we do have some coverage there but not against master.

PR #283, opened 2018-09-28
Add random values in ITs

### LINKED ISSUE #282: Use random values in ITs
As discussed in #258, we should use guids in test account names/emails so that the IT suite can run in parallel on the same org.

laura-rodriguez:
Resolves #282

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/ApplicationScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/ApplicationScenarios.cs
@@ -19,10 +19,11 @@ namespace Okta.Sdk.IntegrationTests
         public async Task AddBookmarkApp()
         {
             var client = TestClient.Create();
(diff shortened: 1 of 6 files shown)


nbarbettini:
Looks like the test failure was unrelated to this change?

vijetmahabaleshwar-okta:
Actually the failure seems like it's because of this PR.
I ran it locally and it's failing there as well.

vijetmahabaleshwar-okta:
@laura-rodriguez - I'm seeing some weird behavior with the logs API.
If I create an app with name = some guid, and if I query logs for the app, I get empty array.
E.g - `/api/v1/logs?q=1caec187-8e05-4d11-a8c1-2617b7df9b2` where `1caec187-8e05-4d11-a8c1-2617b7df9b2` is the app name, I get `[]` back.
Trying to see why this is happening.

PR #293, opened 2018-12-13
Add retry strategy

laura-rodriguez:


src/Okta.Sdk/DefaultRetryStrategy.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DefaultRetryStrategy"/> class.
+        /// </summary>
+        public DefaultRetryStrategy()
+        {
+        }
+
+        /// <inheritdoc/>
+        public async Task<HttpResponseMessage> WaitAndRetryAsync(HttpRequestMessage request, Func<HttpRequestMessage, Task<HttpResponseMessage>> operation)
```
bradhaverstein-okta: Because `DefaultRequestExecutor.SendAsync(...)` accepts a `CancellationToken`, this method also needs to accept a `CancellationToken` and pass it to `CreateDelayAsync` so it can be plumbed through to `Task.Delay`.

Otherwise, when the cancellation token is triggered we won't actually cancel the operation until the current `Task.Delay` completes. It's possible that we'll spend more time in `Task.Delay` than in `operation()` so this is significant.

src/Okta.Sdk/DefaultRetryStrategy.cs
```diff
@@ -0,0 +1,134 @@
...
+                var statusCode = (int)response.StatusCode;
+
+                try
+                {
+                    if (statusCode == 429 && attempts < MaxRetries + 1 &&
+                        ((RequestTimeOut > 0 && stopwatch.Elapsed.Seconds < RequestTimeOut) || RequestTimeOut == 0))
+                    {
+                        await CreateDelayAsync(response).ConfigureAwait(false);
+                        response.Headers.TryGetValues("X-Okta-Request-Id", out var requestId);
+                        AddRetryOktaHeaders(request, requestId.First(), attempts);
```
bradhaverstein-okta: This call to `requestId.First()` is going to throw if the `X-Okta-Request-Id` header wasn't present in the response. That can be a problem if, say, the request failed due to a network timeout or was rejected by Okta before it actually reached monolith (like if Nginx or an AWS network appliance fails with a 502).

We probably need to change it to `FirstOrDefault()` and handle an empty value in `AddRetryOktaHeaders`.

src/Okta.Sdk/DefaultRetryStrategy.cs
```diff
@@ -0,0 +1,134 @@
...
+                    }
+                    else
+                    {
+                        return response;
+                    }
+                }
+                catch
```
bradhaverstein-okta: I don't think I understand the purpose of this `try...catch`. The code inside this block all looks like internal logic and I don't see any situations where it is expected to throw. So if it *did* throw it would be a bug and we definitely would not want to swallow the exception. Also there are exceptions like `OperationCanceledException` or `OutOfMemoryException` that should never be swallowed.

It seems like we should just remove the `try...catch` unless I'm missing something.

src/Okta.Sdk/DefaultRetryStrategy.cs
laura-rodriguez: Good call 👍

src/Okta.Sdk/DefaultRetryStrategy.cs
laura-rodriguez: I throw an `InvalidOperationException`  when we have multiple `X-Rate-Limit-Reset` headers. 
For those case, we shouldn't retry and just return the error.

src/Okta.Sdk/Configuration/OktaClientConfiguration.cs
```diff
@@ -29,6 +39,22 @@ public sealed class OktaClientConfiguration : IDeepCloneable<OktaClientConfigura
         /// </value>
         public int? ConnectionTimeout { get; set; } = DefaultConnectionTimeout;
 
+        /// <summary>
+        /// Gets or sets the time to wait for resolved client request (includes retries), less or equal than 0 means "no timeout"
+        /// </summary>
+        /// <value>
+        /// The request timeout in seconds.
+        /// </value>
+        public int? RequestTimeout { get; set; } = DefaultRequestTimeOut;
```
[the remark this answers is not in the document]
laura-rodriguez: As these values are grabbed from the configuration files we need to consider that we can have missing params.

src/Okta.Sdk/Configuration/OktaClientConfiguration.cs
```diff
@@ -29,6 +39,22 @@ public sealed class OktaClientConfiguration : IDeepCloneable<OktaClientConfigura
...
+
+        /// <summary>
+        /// Gets or sets the number of times to retry
+        /// </summary>
+        /// <value>
+        /// The number of times to retry
+        /// </value>
+        public int? MaxRetries { get; set; } = DefaultMaxRetries;
```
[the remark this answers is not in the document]
laura-rodriguez: Same reason, this value is gonna be grabbed from the config.

src/Okta.Sdk/Configuration/OktaClientConfiguration.cs
```diff
@@ -21,6 +21,16 @@ public sealed class OktaClientConfiguration : IDeepCloneable<OktaClientConfigura
...
 
+        /// <summary>
+        /// The default number of times to retry
+        /// </summary>
+        public static readonly int DefaultMaxRetries = 2;
+
+        /// <summary>
+        /// The default request timeout in seconds
+        /// </summary>
+        public static readonly int DefaultRequestTimeOut = 0;
```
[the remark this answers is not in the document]
laura-rodriguez: They are public in case devs want to use them in their own retry implementations.

src/Okta.Sdk/DefaultRetryStrategy.cs
```diff
@@ -58,39 +62,54 @@ public async Task<HttpResponseMessage> WaitAndRetryAsync(HttpRequestMessage requ
...
             }
             while (true);
         }
 
+        /// <summary>
+        /// Checks if a http response message should be retried
+        /// </summary>
+        /// <param name="response">The http response message</param>
+        /// <returns>True if the value is must be retried, otherwise false.</returns>
+        public virtual bool IsRetryable(HttpResponseMessage response)
```
bradhaverstein-okta: nit: This method could be shortened to one line.
```cs
public virtual bool IsRetryable(HttpResponseMessage response) =>
  response != null && _retryableStatusCodes.Contains(response.StatusCode);
```

src/Okta.Sdk/DefaultRetryStrategy.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// The default delta used in the back-off formula to account for some clock skew in our service
+        /// </summary>
+        public const int DefaultBackOffSecondsDelta = 1;
+
+        private readonly int _maxRetries;
+        private readonly int _requestTimeout;
+        private readonly int _backoffSecondsDelta;
+
+        // Now we are only managing 429 errors, but we can accept other codes in the future
+        private IList<HttpStatusCode> _retryableStatusCodes = new List<HttpStatusCode> { (HttpStatusCode)429 };
```
bradhaverstein-okta: Just a suggestion, but maybe the status codes should be settable in the constructor, with 429 being the default value (just like `backoffSecondsDelta`).

src/Okta.Sdk/DefaultRetryStrategy.cs
```diff
@@ -0,0 +1,158 @@
...
+{
+    /// <summary>
+    /// Default retry strategy
+    /// </summary>
+    public class DefaultRetryStrategy : IRetryStrategy
+    {
+        /// <summary>
+        /// The default delta used in the back-off formula to account for some clock skew in our service
+        /// </summary>
+        public const int DefaultBackOffSecondsDelta = 1;
```
bradhaverstein-okta: In this variable we capitalize backoff as two separate words, but in the private variable we capitalize it like its all one word. They are both defensible but we should be consistent.

The [only example I see in the .NET framework](https://docs.microsoft.com/en-us/dotnet/api/microsoft.exchange.webservices.data.serverbusyexception.backoffmilliseconds?view=exchange-ews-api#Microsoft_Exchange_WebServices_Data_ServerBusyException_BackOffMilliseconds) uses "BackOff" so maybe we should go with that.

src/Okta.Sdk/DefaultRetryStrategy.cs
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            return isRetryable;
+        }
+
+        private void AddRetryOktaHeaders(HttpRequestMessage request, string requestId, int numberOfRetries)
+        {
+            if (!request.Headers.Contains("X-Okta-Retry-For"))
+            {
+                request.Headers.Add("X-Okta-Retry-For", requestId);
```
bradhaverstein-okta: nit: If `requestId` is `null` here what happens? Hopefully we don't get an exception or print an empty `X-Okta-Retry-For` header in the request.

src/Okta.Sdk/DefaultRetryStrategy.cs
bradhaverstein-okta: If what you want is to return the original `response` in this case (and also if the backoff time exceeds the request timeout), I think it would be good to write the code in a way that doesn't require the SDK to throw and immediately catch an exception.

Throwing an exception takes a lot of system resources, so its best practice not to throw an exception and immediately catch it just to implement control flow. It's also confusing because people don't expect a class to throw an exception in one method then catch it in another, and will automatically assume the exceptions are being thrown to the caller (I made this assumption until you explained the logic).

Also there are cases where .NET library methods throw `InvalidOperationException` and you would not want to accidentally catch those.

Maybe rewrite `CreateDelayAsync` so that it returns a `TimeSpan` instead of a Task, and if it returns `TimeSpan.Zero` then you break out of the loop.

bradhaverstein-okta: I'm also wondering why we stop retrying if we see two or more `X-Rate-Limit-Reset` headers. It seems like there are more graceful options for handling that situation:
1. Ignore the `X-Rate-Limit-Reset` header and fallback to using the logic that applies when that header isn't present.
2. Choose one of the values to use (i.e. always use the first one).

src/Okta.Sdk/DefaultRetryStrategy.cs
laura-rodriguez: As we already discussed,  now we only want to support 429.

PR #297, opened 2018-12-31
Adds safe generation of user-agent string

### LINKED ISSUE #296: AmbiguousMatchException: Multiple custom attributes of the same type found
I'm experiencing a runtime issue with this library and I was hoping someone might be able to provide some guidance.  It looks to be caused by [this line](https://github.com/okta/okta-sdk-dotnet/blob/e155419fe2097b0bc490e366e182d173526b58ac/src/Okta.Sdk/Internal/UserAgentBuilder.cs#L44) in the `UserAgentBuilder` class.  I'm running an ASP.NET (not Core) application on .NET Framework.  The runtime exception isn't an issue when running on .NET Core.

I'm not able to reproduce the issue on my Windows 10 machine, but in our test & production environments (Windows Server 2012) we do experience the runtime exception.  I tried bumping up the version of .NET Framework on our test server (to match my development environment), but no dice.

I realize this isn't exactly an issue with the Okta library and it's likely a netstandard vs netframework topic, but I suspect someone else may run into this issue as well.  I've already downgraded one application to the older `Okta.Core.Client` library to address this issue.  I'd like to keep using this library, and unfortunately the issue is caused by the user agent telemetry data being collected which isn't relevant to what I'm trying to do. 

```
System.Reflection.AmbiguousMatchException: Multiple custom attributes of the same type found.
   at System.Attribute.GetCustomAttribute(Assembly element, Type attributeType, Boolean inherit)
   at System.Runtime.InteropServices.RuntimeInformation.get_FrameworkDescription()
   at Okta.Sdk.Internal.UserAgentBuilder.Generate()
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
```

justinhelgerson:
Fixes #296 

Here's my first PR to this project.  🎆 I'll sign the CLA when I'm in the office tomorrow.

A note on the code:  I added an optional `ILogger` in the event someone does want to troubleshoot why the user-agent string is empty.  This way it at least provides a way of digging into it more if they desire. 
 Not sure if I got the message level right though.

### DIFF
--- a/src/Okta.Sdk/Internal/UserAgentBuilder.cs
+++ b/src/Okta.Sdk/Internal/UserAgentBuilder.cs
@@ -43,26 +43,43 @@ namespace Okta.Sdk.Internal
 
         private string Generate()
         {
+            string sdkToken = string.Empty;
+            string runtimeToken = string.Empty;
+            string operatingSystemToken = string.Empty;
+
             try
             {
-                var sdkToken = $"{OktaSdkUserAgentName}/{GetSdkVersion()}";
-
-                var runtimeToken = $"runtime/{Sanitize(RuntimeInformation.FrameworkDescription)}";
-
-                var operatingSystemToken = $"os/{Sanitize(RuntimeInformation.OSDescription)}";
-
-                return string.Join(
-                    " ",
-                    sdkToken,
-                    runtimeToken,
-                    operatingSystemToken)
-                .Trim();
+                sdkToken = $"{OktaSdkUserAgentName}/{GetSdkVersion()}";
             }
             catch (Exception ex)
             {
-                _logger.LogDebug($"An exception was thrown while generating the user agent string.  Exception: {ex.Message}");
-                return string.Empty;
+                _logger.LogDebug($"An error occurred generating the {nameof(sdkToken)} portion of the user-agent string.  Exception: {ex.Message}");
             }
+
+            try
+            {
+                runtimeToken = $"runtime/{Sanitize(RuntimeInformation.FrameworkDescription)}";
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug($"An error occurred generating the {nameof(runtimeToken)} portion of the user-agent string.  Exception: {ex.Message}");
+            }
+
+            try
+            {
+                operatingSystemToken = $"os/{Sanitize(RuntimeInformation.OSDescription)}";
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug($"An error occurred generating the {nameof(operatingSystemToken)} portion of the user-agent string.  Exception: {ex.Message}");
+            }
+
+            return string.Join(
+                " ",
+                sdkToken,
+                runtimeToken,
+                operatingSystemToken)
+            .Trim();
         }
 
         private static string GetSdkVersion()
--- a/src/Okta.Sdk/Internal/UserAgentBuilder.cs
+++ b/src/Okta.Sdk/Internal/UserAgentBuilder.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 
 namespace Okta.Sdk.Internal
 {
@@ -31,7 +32,7 @@ namespace Okta.Sdk.Internal
         /// <param name="logger">An optional logging interface</param>
         public UserAgentBuilder(ILogger logger = null)
         {
-            _logger = logger;
+            _logger = logger ?? NullLogger.Instance;
             _cachedUserAgent = new Lazy<string>(Generate);
         }
 
@@ -53,7 +54,7 @@ namespace Okta.Sdk.Internal
             }
             catch (Exception ex)
             {
-                _logger.LogDebug($"An error occurred generating the {nameof(sdkToken)} portion of the user-agent string.  Exception: {ex.Message}");
+                _logger.LogError($"An error occurred generating the {nameof(sdkToken)} portion of the user-agent string.  Exception: {ex.Message}");
             }
 
             try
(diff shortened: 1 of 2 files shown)


nbarbettini:
A debug loglevel makes sense to me.

Since we know that the `RuntimeInformation.FrameworkDescription` line could break (but `GetSdkVersion()` is less likely to), want to split this up into three try/catch, one for each "chunk" of the string?

justinhelgerson:
@nbarbettini I ran the unit tests locally and everything checks out.  I was going to put together a unit test for my change but mocking the `RuntimeInformation` might be a pain in the arse.

I added try/catch around the individual parts like you suggested.  Let me know if you have any other feedback.

nbarbettini:
@justinhelgerson Yup, mocking that doesn't sound fun. I don't think it's necessary.

This looks good to me! @laura-rodriguez is out on holiday right now, so I will get some other review eyes on it and then merge. Thanks again for contributing!

src/Okta.Sdk/Internal/UserAgentBuilder.cs
```diff
@@ -22,12 +23,15 @@ public sealed class UserAgentBuilder
...
+        private readonly ILogger _logger;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="UserAgentBuilder"/> class.
         /// </summary>
-        public UserAgentBuilder()
+        /// <param name="logger">An optional logging interface</param>
+        public UserAgentBuilder(ILogger logger = null)
         {
+            _logger = logger;
```
bradhaverstein-okta: Here we are defaulting `_logger` to null, but the code below is going to throw a `NullReferenceException` in that case.

Technically you can fix this by using the null coalescing operator everywhere. But it's very easy to forget one place, and most of the time unit tests don't exhaustively cover logging.
```cs
_logger?.LogDebug("An error occurred.");
```

Instead I would recommend initializing `_logger` to a [NullLogger](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.abstractions.nulllogger?view=aspnetcore-2.1) if the `logger` parameter is null.

src/Okta.Sdk/Internal/UserAgentBuilder.cs
```diff
@@ -22,12 +23,15 @@ public sealed class UserAgentBuilder
...
         // Lazy ensures this only runs once and is cached.
         private readonly Lazy<string> _cachedUserAgent;
+        private readonly ILogger _logger;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="UserAgentBuilder"/> class.
         /// </summary>
-        public UserAgentBuilder()
+        /// <param name="logger">An optional logging interface</param>
+        public UserAgentBuilder(ILogger logger = null)
```
bradhaverstein-okta: The typical pattern is for classes to accept an `ILogger` whose type parameter is the class itself. That way the logging framework can automatically trace the name of the class along with any log messages.
```cs
public UserAgentBuilder(ILogger<UserAgentBuilder> logger)
{
  ...
}
```

If you are using the `Microsoft.Extensions.DependencyInjection` framework the correct kind of logger can be automatically created and passed to the constructor. If you aren't, you can use an [ILoggerFactory](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.iloggerfactory?view=aspnetcore-2.1) to create the right kind of logger by calling [CreateLogger&lt;T&gt;()](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerfactoryextensions.createlogger?view=aspnetcore-2.1#Microsoft_Extensions_Logging_LoggerFactoryExtensions_CreateLogger__1_Microsoft_Extensions_Logging_ILoggerFactory_).

src/Okta.Sdk/Internal/UserAgentBuilder.cs
```diff
@@ -39,11 +43,36 @@ public UserAgentBuilder()
...
+            string operatingSystemToken = string.Empty;
 
-            var runtimeToken = $"runtime/{Sanitize(RuntimeInformation.FrameworkDescription)}";
+            try
+            {
+                sdkToken = $"{OktaSdkUserAgentName}/{GetSdkVersion()}";
+            }
+            catch (Exception ex)
+            {
+                _logger.LogDebug($"An error occurred generating the {nameof(sdkToken)} portion of the user-agent string.  Exception: {ex.Message}");
```
bradhaverstein-okta: I think you should probably go with `LogError` instead of `LogDebug` unless you are really expecting this to fail often in common scenarios. Messages logged with `LogDebug` are considered very verbose and are often not even displayed or recorded unless the logging level is changed from its default.

src/Okta.Sdk/Internal/UserAgentBuilder.cs
justinhelgerson: Yep, good point.  I switched it to use an instance of `NullLogger` if needed.

src/Okta.Sdk/Internal/UserAgentBuilder.cs
justinhelgerson: We could flip to `ILogger<T>`, but `NullLogger<T>` isn't available until (I think) 2.1?  So I skipped this for now.

src/Okta.Sdk/Internal/UserAgentBuilder.cs
justinhelgerson: > Messages logged with LogDebug are considered very verbose and are often not even displayed or recorded unless the logging level is changed from its default.

That's why I chose debug.  I don't think it makes sense for people to see this error unless they're digging deep.  The 99% just want their API call to work.  Seeing this logged as an error seems confusing when the function performed its purpose (i.e., to the developer, there wasn't an error).  My two cents.

bradhaverstein-okta: Hmmm, OK that makes sense to me. Thanks!

src/Okta.Sdk/Internal/UserAgentBuilder.cs
bradhaverstein-okta: Yeah I see other places in the code where `ILogger` is being used instead of `ILogger<T>`, so this is probably something the devex team should consider changing. I guess that goes a bit beyond the scope of your change.

I'm not on the devex team, BTW, just helping with .NET reviews. 😄

nbarbettini:
Thanks again @justinhelgerson. ❤️ 

Please open more issues if you run into anything else!

PR #300, opened 2019-01-11
Remove connection: close header from DefaultHttpClient.

laura-rodriguez:
Fixes [OKTA-202670](https://oktainc.atlassian.net/browse/OKTA-202670)

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/DefaultDataStoreShould.cs
+++ b/src/Okta.Sdk.IntegrationTests/DefaultDataStoreShould.cs
@@ -34,7 +34,7 @@ namespace Okta.Sdk.IntegrationTests
             }, null, default(CancellationToken));
 
             var connectionHeader = response.Headers.FirstOrDefault(x => x.Key.Equals("connection", StringComparison.OrdinalIgnoreCase));
-            connectionHeader.Value.Any(s => s.Equals("close", StringComparison.OrdinalIgnoreCase)).Should().BeFalse();
+            connectionHeader.Value?.Any(s => s.Equals("close", StringComparison.OrdinalIgnoreCase)).Should().BeFalse();
         }
     }
 }
--- a/src/Okta.Sdk.IntegrationTests/DefaultDataStoreShould.cs
+++ b/src/Okta.Sdk.IntegrationTests/DefaultDataStoreShould.cs
@@ -1,12 +1,14 @@
-﻿using System;
-using System.Collections.Generic;
+﻿// <copyright file="DefaultDataStoreShould.cs" company="Okta, Inc">
+// Copyright (c) 2014 - present Okta, Inc. All rights reserved.
+// Licensed under the Apache 2.0 license. See the LICENSE file in the project root for full license information.
+// </copyright>
+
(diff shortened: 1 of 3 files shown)


nbarbettini:
LGTM. Some background info: I originally added `Connection: close` as an attempt to solve [this problem](https://github.com/dotnet/corefx/issues/11224), but it was a premature optimization because I did not actually _see_ that problem. I was just reading too much about the .NET HTTP stack and thought it was a clever fix. The consensus from that thread seems to be that it is a very unlikely problem to run into.

src/Okta.Sdk.IntegrationTests/DefaultDataStoreShould.cs
```diff
@@ -0,0 +1,40 @@
...
+            var requestExecutor = new DefaultRequestExecutor(configuration, httpClient, NullLogger.Instance);
+            var resourceFactory = new ResourceFactory(testOktaClient, NullLogger.Instance);
+            var dataStore = new DefaultDataStore(requestExecutor, new DefaultSerializer(), resourceFactory, NullLogger.Instance);
+
+            var response = await dataStore.GetAsync<Resource>(
+                new HttpRequest
+            {
+                // Endpoint that returns a chunked response
+                Uri = $"/api/v1/meta/schemas/user/default",
+            }, null, default(CancellationToken));
```
bradhaverstein-okta: nit: The indentation here is a little off. Lines 31-34 should probably be indented to align with line 30.

Another option is to initialize the `HttpRequest` as a separate variable on another line.

src/Okta.Sdk/Internal/DefaultHttpClient.cs
```diff
@@ -42,10 +45,8 @@ public static class DefaultHttpClient
             {
                 Timeout = TimeSpan.FromSeconds(connectionTimeout ?? OktaClientConfiguration.DefaultConnectionTimeout),
             };
-            logger.LogTrace($"Using timeout of {client.Timeout} second(s) from configuration");
 
-            // Workaround for https://github.com/dotnet/corefx/issues/11224
-            client.DefaultRequestHeaders.Add("Connection", "close");
```
bradhaverstein-okta: Do we have a bug tracking the fact that we need to come up with some other workaround to ensure that the `HttpClient` in the SDK switches to the new server in a timely manner when DNS is updated for a failover?

Also, my reading of [the .NET Core thread on HttpClient failover](https://github.com/dotnet/corefx/issues/11224) is that the `connection: close` header should be sent by the *server*, not the client. I think what Darrell Miller is suggesting is that Okta's API servers in the failed region could begin issuing `connection: close` headers to all clients during a failover to ensure that the clients switch to the new endpoint. In other words, the server can close the HTTP connection instead of assuming that the client will be well behaved.

So that might explain why sending `connection: close` from the client is breaking things. 😄

laura-rodriguez: I've created an issue on GitHub so we can prioritize later. Thanks @bradhaverstein-okta!

Issue #309, opened 2019-03-14
Add support for SMS Auto-Activate


deano-au:
Add support for SMS Auto-Activate as per:
[https://developer.okta.com/docs/api/resources/factors#enroll-and-auto-activate-okta-sms-factor](https://developer.okta.com/docs/api/resources/factors#enroll-and-auto-activate-okta-sms-factor)

laura-rodriguez:
Hi @deano-au,

Sorry for the late response. We already support this use case, you have to use the following code:

```
 await user.AddFactorAsync(new AddSmsFactorOptions()
                {
                    PhoneNumber = "{userPhoneNumber}",
                });
```
I've created an[ integration test](https://github.com/okta/okta-sdk-dotnet/pull/317/files#diff-29df9fc0b58f5d24b8b8a11b7684ab54R62.) to make sure that this is working properly, you can check out there as well. 
Let me know if this helps. Thanks!

PR #317, opened 2019-05-15
Add an IT for enroll SMS use case

laura-rodriguez:
People asking about this in #309 so I added a test to make sure it's working.

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/FactorScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/FactorScenarios.cs
@@ -15,7 +15,7 @@ namespace Okta.Sdk.IntegrationTests
     public class FactorScenarios
     {
         [Fact]
-        public async Task CreateSecurityQuestionFactor()
+        public async Task EnrollSecurityQuestionFactor()
         {
             var client = TestClient.Create();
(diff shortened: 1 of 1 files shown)


src/Okta.Sdk.IntegrationTests/FactorScenarios.cs
```diff
@@ -58,6 +58,48 @@ public async Task CreateSecurityQuestionFactor()
...
+            {
+                Profile = profile,
+                Password = "Abcd1234",
+            });
+
+            try
+            {
+                await createdUser.AddFactorAsync(new AddSmsFactorOptions()
+                {
+                    PhoneNumber = "+15554151337",
```
vijetmahabaleshwar-okta: What phone number is this? 
For node and java SDK, we use a google voice number - https://github.com/okta/okta-sdk-nodejs/blob/master/test/it/factor-create.js#L87

laura-rodriguez: I grabbed this number from [docs](https://developer.okta.com/docs/api/resources/factors/#enroll-okta-sms-factor). We don't need a real number for this specific test. Besides the consistency, is there any other reason to use that?

PR #324, opened 2019-07-11
Do not skip empty arrays in serialization

laura-rodriguez:
Fixes OKTA-229987.

Okta .NET SDK: UpdateAsync() does not add the empty array profile attributes, leading to profile update failures when attributes are mastered by another IdP.


laura-rodriguez:
> nice!
> I'm assuming everything uses that `ResourceSerializingConverter` (Groups, etc), so +1

Yes, all of them are `Resources` so they'll use the same serializer.

PR #332, opened 2019-07-26
Update README and module's manifest indicating the deprecation.

laura-rodriguez:
I still need to verify that the description update is enough for PowerShell Gallery to update the description of the module or if a new patch is required as we had to do with NuGet and the legacy .NET SDK. I'll figure this out when I publish the updated `Okta.Core.Automation.psd1`.


OkGoDoIt:
Do we maybe want to update the changlog or provide any additional details in the readme?  Maybe calling out why it's being deprecated and the recommended path for people who had been using it?

laura-rodriguez:
I confirm that I had to update the manifest's version to publish an updated psd1.

PR #334, opened 2019-08-09
Modifies OktaClient constructor to set a custom JSON Serializer.

laura-rodriguez:
* Modifies DefaultSerializer constructor to set custom settings.
* Fixes [OKTA-242031]
* Some indentation changes for cleanup

### DIFF
--- a/src/Okta.Sdk.UnitTests/DefaultDataStoreShould.cs
+++ b/src/Okta.Sdk.UnitTests/DefaultDataStoreShould.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Microsoft.Extensions.Logging.Abstractions;
+using Newtonsoft.Json;
 using NSubstitute;
 using Okta.Sdk.Internal;
 using Okta.Sdk.UnitTests.Internal;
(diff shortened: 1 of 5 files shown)


src/Okta.Sdk.UnitTests/DefaultDataStoreShould.cs
```diff
@@ -118,6 +119,51 @@ public async Task DelegatePostToRequestExecutor()
...
+            response.Payload.Profile["startDate"].GetType().Should().Be(typeof(string));
+            response.Payload.Profile["startDate"].Should().Be("2016-06-07T00:00:00.000");
+
+            // DateParseHandling.DateTimeOffset: Date formatted strings, e.g. "\/Date(1198908717056)\/" and "2012-03-21T05:40Z", are parsed to DateTimeOffset.
+            dataStore = new DefaultDataStore(mockRequestExecutor, new DefaultSerializer(new JsonSerializerSettings() { DateParseHandling = DateParseHandling.DateTimeOffset }), new ResourceFactory(null, null), NullLogger.Instance);
+            response = await dataStore.GetAsync<User>(request, new RequestContext(), CancellationToken.None);
+            response.Payload.Profile["startDate"].GetType().Should().Be(typeof(DateTimeOffset));
+            var dateTimeOffset = DateTimeOffset.Parse("2016-06-07T00:00:00.000");
+            response.Payload.Profile["startDate"].Should().Be(dateTimeOffset);
+        }
```
OkGoDoIt: I’m not really sure what `ContractResolver` is in the settings, but does it need to be tested?

laura-rodriguez: not sure if I follow, what `ContractResolver`?

laura-rodriguez:
> Looks like there’s quite a bit of duplicate code in the 2 OktaClient constructors, is that something we should clean up? Maybe not relevant to this PR, so not a blocker.
> 
> Do we need to test for what happens if a `ContractResolver` is set or someone passes in some other implementation of `ISerializer`? I assume those are very much edge cases but I don’t know how it affects things.
> 
> Anyway those are all just nits, this PR looks good to me.

IMO If devs decide to implement their own Serializer they are responsible to make the SDK still works as expected. In the meantime, they can always use the `DefaultSerializer` and customize it by passing a `JsonSerializerSettings`.

src/Okta.Sdk/Internal/DefaultSerializer.cs
```diff
@@ -23,14 +23,22 @@ public sealed class DefaultSerializer : ISerializer
...
-            };
+                _serializer = JsonSerializer.Create(settings);
+            }
+            else
+            {
+                _serializer = new JsonSerializer
+                {
+                    NullValueHandling = NullValueHandling.Ignore,
+                    DefaultValueHandling = DefaultValueHandling.Ignore,
+                    ContractResolver = new DefaultContractResolver(),
```
OkGoDoIt: This ContractResolver. We normally use a DefaultContractResolver but if a dev passes in a custom settings then it will use whatever is in there, not sure if it has the same default of if it defaults to null or what, or frankly what it even does in the first place. I doubt it’s an issue, just wanted to see if you thought it’s worth looking into.

PR #345, opened 2019-09-18
Add null checking to `ExtractHeaders`

### LINKED ISSUE #344: Exception getting thrown during high load scenarios
When we query a large number of records, we occasionally get an exception thrown. I think the response might not have any `Content` which is causing the NullReferenceException. It would probably be good to check `response.Content != null` before performing the Concat.

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Okta.Sdk.Internal.DefaultRequestExecutor.ExtractHeaders(HttpResponseMessage response)
   at Okta.Sdk.Internal.DefaultRequestExecutor.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Okta.Sdk.Internal.DefaultDataStore.GetAsync[T](HttpRequest request, RequestContext requestContext, CancellationToken cancellationToken)
   at Okta.Sdk.OktaClient.GetAsync[T](HttpRequest request, CancellationToken cancellationToken)
   at Okta.Sdk.UsersClient.GetUserAsync(String userId, CancellationToken cancellationToken)
```

cwoolum:
This PR adds null checking to the response headers. This should fix #344.

### DIFF
--- a/src/Okta.Sdk/Internal/DefaultRequestExecutor.cs
+++ b/src/Okta.Sdk/Internal/DefaultRequestExecutor.cs
@@ -116,7 +116,16 @@ namespace Okta.Sdk.Internal
         }
 
         private static IEnumerable<KeyValuePair<string, IEnumerable<string>>> ExtractHeaders(HttpResponseMessage response)
-            => response.Headers.Concat(response.Content.Headers);
+        {
+            var responseContentHeaders = response.Content?.Headers;
+
+            if (responseContentHeaders == null)
+            {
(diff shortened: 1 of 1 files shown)


laura-rodriguez:
@cwoolum thanks for your contribution ❤️ !
I'm working on other .NET projects now, but I will take care of your PR as soon as I can :).

laura-rodriguez:
Closing since it is no longer applicable. It seems that null checking was already added in previous versions.

PR #352, opened 2019-11-07
Rate Limit preservation functionality

phi1ipp:
We try to avoid rate limit violations to keep that system log functionality as an indication of a possible problem with applications or security, so this change allows applications to throttle their API call frequency by preserving a certain percent of API rate.

It can also be useful for bulk loading/reporting operations to not affect real applications by consuming all limit.

### DIFF
--- a/src/Okta.Sdk/Configuration/OktaClientConfiguration.cs
+++ b/src/Okta.Sdk/Configuration/OktaClientConfiguration.cs
@@ -84,6 +84,15 @@ namespace Okta.Sdk.Configuration
         public string Token { get; set; }
 
         /// <summary>
+        /// Get or sets Rate Limit preservation percent
+        /// </summary>
+        /// <value>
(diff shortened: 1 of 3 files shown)


laura-rodriguez:
Hi @phi1ipp ,

Thanks for your contribution ❤️ ! 
Since this seems a new feature, I need to discuss it with the team first, because we want to be consistent across all our SDKs (including other languages). I'll let you know when I have updates.
Thanks again!! 🥇

laura-rodriguez:
Hi @phi1ipp ,

Apologize we haven't followed up yet.

Finally, we discussed this feature with the team, and we decided to keep the built-in retry strategy as simple as possible. However, when designing this feature we took into account that customers might want to provide their own rate-limiting logic, that's why we provide a default behavior and we also support custom retry implementations via DI. 

Thank you so much for your time and your contribution!

PR #358, opened 2020-01-23
Multitarget .NET frameworks

### LINKED ISSUE #348: IAsyncEnumerable issues with .net core 3.0
Hello, I'm trying to use Okta Sdk with .net core 3.0 and I also have my custom IAsyncEnumerables in the code. Unfortunately it doesn't compile as Okta Sdk uses System.Interactive.Async version 3.1.1, but .net core 3.0 already includes IAsyncEnumberable into bcl with different signature - it's a known breaking change. Here is a quick example:

```csharp
    class Program
    {
        static void Main(string[] args)
        {
            var client = new OktaClient(new OktaClientConfiguration
            {
                OktaDomain = "https://{{yourOktaDomain}}",
                Token = "{{yourApiToken}}"
            });

            client.Policies.ListPolicies("").ToList();
        }
        
        public class MyAsync : IAsyncEnumerable<string>
        {
            public ValueTask<bool> MoveNextAsync() => default;
            public ValueTask DisposeAsync() => default;
        }
    }
```

This code fails to compile with the following error:

`Program.cs(24,32): error CS0433: The type 'IAsyncEnumerable<T>' exists in both 'System.Interactive.Async, Version=3.0.3000.0, Culture=neutral, PublicKeyToken=94bc3704cddfc263' and 'System.Runtime, Version=4.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'`

Any chance to add IAsyncEnumerable support from .net core 3.0/netstandard2.1? It's possible to multitarget Okta.Sdk library for example.

laura-rodriguez:
* Upgrade .NET Standard version to 2.0. 
* Add support for .NET Core 3.x via multitarget.
* Upgrade NuGet dependencies.
* Fix #348 
* Fix [OKTA-277647](https://oktainc.atlassian.net/browse/OKTA-277647)
* Also, in this PR I disabled [StyleCop rule # 1629](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1629.md) since it doesn't add any value.

### DIFF
--- a/src/Okta.Sdk/ApplicationsClient.cs
+++ b/src/Okta.Sdk/ApplicationsClient.cs
@@ -14,7 +14,7 @@ namespace Okta.Sdk
     public sealed partial class ApplicationsClient : OktaClient, IApplicationsClient, IAsyncEnumerable<IApplication>
     {
         /// <inheritdoc/>
-        public IAsyncEnumerator<IApplication> GetEnumerator() => ListApplications().GetEnumerator();
+       // public IAsyncEnumerator<IApplication> GetEnumerator() => ListApplications().GetEnumerator();
 
         /// <inheritdoc/>
         public async Task<T> GetApplicationAsync<T>(string appId, CancellationToken cancellationToken = default(CancellationToken))
@@ -427,5 +427,10 @@ namespace Okta.Sdk
         /// <inheritdoc />
         public async Task DeleteApplicationUserAsync(string appId, string userId)
             => await DeleteApplicationUserAsync(appId, userId, false, default(CancellationToken)).ConfigureAwait(false);
+
+        public IAsyncEnumerator<IApplication> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+        {
+            return ListApplications().GetAsyncEnumerator();
+        }
     }
 }
--- a/src/Okta.Sdk/ApplicationsClient.cs
+++ b/src/Okta.Sdk/ApplicationsClient.cs
@@ -14,7 +14,7 @@ namespace Okta.Sdk
     public sealed partial class ApplicationsClient : OktaClient, IApplicationsClient, IAsyncEnumerable<IApplication>
     {
         /// <inheritdoc/>
-       // public IAsyncEnumerator<IApplication> GetEnumerator() => ListApplications().GetEnumerator();
+        public IAsyncEnumerator<IApplication> GetAsyncEnumerator(CancellationToken cancellationToken = default) => ListApplications().GetAsyncEnumerator(cancellationToken);
 
         /// <inheritdoc/>
         public async Task<T> GetApplicationAsync<T>(string appId, CancellationToken cancellationToken = default(CancellationToken))
@@ -427,10 +427,5 @@ namespace Okta.Sdk
         /// <inheritdoc />
         public async Task DeleteApplicationUserAsync(string appId, string userId)
             => await DeleteApplicationUserAsync(appId, userId, false, default(CancellationToken)).ConfigureAwait(false);
-
-        public IAsyncEnumerator<IApplication> GetAsyncEnumerator(CancellationToken cancellationToken = default)
-        {
-            return ListApplications().GetAsyncEnumerator();
-        }
     }
 }
--- a/src/Okta.Sdk/CollectionClient{T}.cs
+++ b/src/Okta.Sdk/CollectionClient{T}.cs
@@ -34,6 +34,7 @@ namespace Okta.Sdk
             _requestContext = requestContext;
         }
 
+        /// <inheritdoc/>
         public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default)
         => new CollectionAsyncEnumerator<T>(_dataStore, _initialRequest, _requestContext);
 
--- a/src/Okta.Sdk/CollectionClient{T}.cs
+++ b/src/Okta.Sdk/CollectionClient{T}.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using Okta.Sdk.Internal;
 
 namespace Okta.Sdk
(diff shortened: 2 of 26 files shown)


src/Okta.Sdk/ApplicationsClient.cs
```diff
@@ -14,7 +14,7 @@ namespace Okta.Sdk
     public sealed partial class ApplicationsClient : OktaClient, IApplicationsClient, IAsyncEnumerable<IApplication>
     {
         /// <inheritdoc/>
-        public IAsyncEnumerator<IApplication> GetEnumerator() => ListApplications().GetEnumerator();
+       // public IAsyncEnumerator<IApplication> GetEnumerator() => ListApplications().GetEnumerator();
```
laura-rodriguez: TODO: Remove commented code.

src/Okta.Sdk/GroupsClient.cs
```diff
@@ -29,7 +29,12 @@ public Task<IGroup> CreateGroupAsync(CreateGroupOptions options, CancellationTok
...
         }
 
+        public IAsyncEnumerator<IGroup> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+        {
+            return ListGroups().GetAsyncEnumerator();
+        }
+
         /// <inheritdoc/>
-        public IAsyncEnumerator<IGroup> GetEnumerator() => ListGroups().GetEnumerator();
+        //public IAsyncEnumerator<IGroup> GetEnumerator() => ListGroups().GetEnumerator();
```
laura-rodriguez: TODO: Remove commented code.

src/Okta.Sdk/LogsClient.cs
```diff
@@ -4,16 +4,22 @@
...
+        {
+            return GetLogs().GetAsyncEnumerator();
+        }
+
         /// <summary>
         /// Gets the LogsClient enumerator
         /// </summary>
         /// <returns>A collection of <see cref="ILogEvent"/> that can be enumerated asynchronously</returns>
-        public IAsyncEnumerator<ILogEvent> GetEnumerator() => GetLogs().GetEnumerator();
+        //public IAsyncEnumerator<ILogEvent> GetEnumerator() => GetLogs().GetEnumerator();
```
laura-rodriguez: TODO: Remove commented code.

src/Okta.Sdk/UsersClient.cs
```diff
@@ -16,7 +16,7 @@ namespace Okta.Sdk
     public sealed partial class UsersClient : OktaClient, IUsersClient, IAsyncEnumerable<IUser>
     {
         /// <inheritdoc/>
-        public IAsyncEnumerator<IUser> GetEnumerator() => ListUsers().GetEnumerator();
+        //public IAsyncEnumerator<IUser> GetEnumerator() => ListUsers().GetEnumerator();
```
laura-rodriguez: TODO: Remove commented code.

src/Okta.Sdk/CollectionClient{T}.cs
```diff
@@ -34,8 +35,8 @@ public sealed class CollectionClient<T> : ICollectionClient<T>
         }
 
         /// <inheritdoc/>
-        public IAsyncEnumerator<T> GetEnumerator()
-            => new CollectionAsyncEnumerator<T>(_dataStore, _initialRequest, _requestContext);
+        public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+        => new CollectionAsyncEnumerator<T>(_dataStore, _initialRequest, _requestContext);
```
bradhaverstein-okta: Here the `cancellationToken` is being thrown away. But it seems like the right fix is to update `CollectionAsyncEnumerator` (and the `PagedCollectionEnumerator` it contains) to accept the cancellation token in their constructors instead of through `MoveNext()`. Is there a bug tracking this work?

It could be that I'm missing something but to me it doesn't seem like that change is a ton of work. Maybe it is worth doing this now?

laura-rodriguez: How to use this token is still unclear to me. The `IAsyncEnumerable` interface forces you to accept a cancellation token in the `GetAsyncEnumerator()` method which is not clear yet how to use it inside of the enumerator:

```
    public interface IAsyncEnumerable<out T>
    {
        IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default);
    }
```
[I created an issue on MS docs repo ](https://github.com/dotnet/docs/issues/17036)to add more details about what the recommended practice is here, since I couldn't find any good content that explains this clearly.

bradhaverstein-okta: I was thinking that we would use it here, in the call to `IDataStore.GetArrayAsync<T>(...)` that `PagedCollectionEnumerator<T>.MoveNextAsync(...)` makes. My interpretation of the interface is that passing the `CancellationToken` to the enumerator's constructor should be equivalent to passing it to any of the enumerator's methods when they are called.

https://github.com/okta/okta-sdk-dotnet/blob/876fb997886a23785d8827d4f637ea6bbde02a14/src/Okta.Sdk/PagedCollectionEnumerator%7BT%7D.cs#L50-L53

laura-rodriguez: Oh I see what you mean, and yes, you're right. This needs to be fixed. Thanks for the heads up.

src/Okta.Sdk/PagedCollectionEnumerator{T}.cs
```diff
@@ -17,30 +17,32 @@ public class PagedCollectionEnumerator<T> : IPagedCollectionEnumerator<T>
     {
         private readonly IDataStore _dataStore;
         private readonly RequestContext _requestContext;
-
         private HttpRequest _nextRequest;
-
+        private CancellationToken _cancellationToken;
```
bradhaverstein-okta: nit: This can be `readonly` I think.

PR #366, opened 2020-03-24
Add workaround for runtime reporting issue when using .NET Core 2.x.

laura-rodriguez:
Fix OKTA-276806

### DIFF
--- a/src/Okta.Sdk/Internal/UserAgentBuilder.cs
+++ b/src/Okta.Sdk/Internal/UserAgentBuilder.cs
@@ -59,7 +59,7 @@ namespace Okta.Sdk.Internal
 
             try
             {
-                runtimeToken = $"runtime/{Sanitize(GetFrameworkDescription())}";
+                runtimeToken = $"runtime/{Sanitize(UserAgentHelper.GetFrameworkDescription())}";
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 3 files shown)


src/Okta.Sdk/Internal/UserAgentBuilder.cs
```diff
@@ -83,6 +83,31 @@ private string Generate()
...
+
+                    if (dotnetCoreAppIndex > 0 && dotnetCoreAppIndex < runtimeAssemblyLocation.Length - 2)
+                    {
+                        frameworkDescription = $".NET Core {runtimeAssemblyLocation[dotnetCoreAppIndex + 1]}";
+                    }
+                }
+            }
+
+            return frameworkDescription;
+        }
```
bradhaverstein-okta: I would recommend adding some unit tests for the string manipulation logic.

Usually I don't spend a lot of time asking for unit tests in code reviews because in many cases they don't really test much. But string manipulation is one of the few examples where unit tests are actually really useful. It's easy to have a bug and also easy to test.

src/Okta.Sdk/Internal/UserAgentBuilder.cs
```diff
@@ -83,6 +83,31 @@ private string Generate()
...
+            if (RuntimeInformation.FrameworkDescription.StartsWith(".NET Core", StringComparison.OrdinalIgnoreCase))
+            {
+                var runtimeAssembly = typeof(System.Runtime.GCSettings).GetTypeInfo().Assembly;
+
+                if (runtimeAssembly != null)
+                {
+                    var runtimeAssemblyLocation = runtimeAssembly.CodeBase.Split(new[] { '/', '\\' }, StringSplitOptions.RemoveEmptyEntries);
+                    int dotnetCoreAppIndex = Array.IndexOf(runtimeAssemblyLocation, "Microsoft.NETCore.App");
+
+                    if (dotnetCoreAppIndex > 0 && dotnetCoreAppIndex < runtimeAssemblyLocation.Length - 2)
```
bradhaverstein-okta: nit: Since `Array.IndexOf(...)` returns `-1` to indicate that the entry doesn't exist should we be considering `dotnetCoreAppIndex = 0` to be an allowed value in this bounds check? Or are we intentionally excluding the case where the path starts with `Microsoft.NETCore.App`?
```suggestion
                    if (dotnetCoreAppIndex >= 0 && dotnetCoreAppIndex < runtimeAssemblyLocation.Length - 2)
```

laura-rodriguez: Good catch!

PR #378, opened 2020-05-15
OKTA-295706 add usertypes

bryanapellanes-okta:
## Issue \#
OKTA-295706

## Code

- [x] Integration test(s)
- [x] Implementation

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/ApplicationScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/ApplicationScenarios.cs
@@ -372,11 +372,12 @@ namespace Okta.Sdk.IntegrationTests
         {
             var client = TestClient.Create();
             var guid = Guid.NewGuid();
+            var testClientId = $"{nameof(AddOpenIdConnectApp)}_TestClientId";
 
             var createdApp = await client.Applications.CreateApplicationAsync(new CreateOpenIdConnectApplication
             {
                 Label = $"dotnet-sdk: AddOpenIdConnectApp {guid}",
-                ClientId = "0oae8mnt9tZcGcMXG0h3",
+                ClientId = testClientId,
                 TokenEndpointAuthMethod = OAuthEndpointAuthenticationMethod.ClientSecretPost,
                 AutoKeyRotation = true,
                 ClientUri = "https://example.com/client",
@@ -413,7 +414,7 @@ namespace Okta.Sdk.IntegrationTests
                 retrieved.Name.Should().Be("oidc_client");
                 retrieved.Label.Should().Be($"dotnet-sdk: AddOpenIdConnectApp {guid}");
                 retrieved.SignOnMode.Should().Be(ApplicationSignOnMode.OpenIdConnect);
-                retrieved.Credentials.OauthClient.ClientId.Should().Be("0oae8mnt9tZcGcMXG0h3");
+                retrieved.Credentials.OauthClient.ClientId.Should().Be(testClientId);
                 retrieved.Credentials.OauthClient.AutoKeyRotation.Should().BeTrue();
                 retrieved.Credentials.OauthClient.TokenEndpointAuthMethod.Should().Be(OAuthEndpointAuthenticationMethod.ClientSecretPost);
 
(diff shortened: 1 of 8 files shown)


src/Okta.Sdk.IntegrationTests/ApplicationScenarios.cs
```diff
@@ -372,11 +372,12 @@ public async Task AddOpenIdConnectApp()
         {
             var client = TestClient.Create();
             var guid = Guid.NewGuid();
+            var testClientId = $"{nameof(AddOpenIdConnectApp)}_TestClientId";
 
             var createdApp = await client.Applications.CreateApplicationAsync(new CreateOpenIdConnectApplication
             {
                 Label = $"dotnet-sdk: AddOpenIdConnectApp {guid}",
-                ClientId = "0oae8mnt9tZcGcMXG0h3",
+                ClientId = testClientId,
```
laura-rodriguez, line 380: Why this change is needed?

bryanapellanes-okta: This was causing integration tests to fail because the same value was being used by multiple tests.  There was a 400 returned by the server because an object with the same clientId already exist.

bryanapellanes-okta: I could move this change into a separate branch/set of changes so usertypes is isolated if you prefer.

laura-rodriguez: oh right! this is the one you mentioned today. It's Ok, we can keep it here.

src/Okta.Sdk.IntegrationTests/UserTypesClientShould.cs
```diff
@@ -0,0 +1,269 @@
...
+                    Description = updatedDescription,
+                    DisplayName = updatedDisplayName,
+                };
+
+                var updatedUserType = await testClient.UserTypes.UpdateUserTypeAsync(updatesToUserType, createdUserType.Id);
+                updatedUserType.Id.Should().Be(createdUserType.Id);
+                updatedUserType.Description.Should().Be(updatedDescription);
+                updatedUserType.DisplayName.Should().Be(updatedDisplayName);
+
+                var retrievedForValidation = await testClient.UserTypes.GetUserTypeAsync(createdUserType.Id);
```
laura-rodriguez, line 119: nit: I suggest using more descriptive names. I would rename this to `retrievedUserTypeForValidation`, so we know it is a user type.

src/Okta.Sdk.IntegrationTests/UserTypesClientShould.cs
```diff
@@ -0,0 +1,269 @@
...
+                        DisplayName = replacedDisplayName,
+                        Name = testName,
+                    }, createdUserType.Id);
+
+                replacedUserType.Id.Should().Be(createdUserType.Id);
+                replacedUserType.Description.Should().Be(replacedDescription);
+                replacedUserType.DisplayName.Should().Be(replacedDisplayName);
+                replacedUserType.Name.Should().Be(testName);
+
+                var retrievedForValidation = await testClient.UserTypes.GetUserTypeAsync(createdUserType.Id);
```
laura-rodriguez, line 165: nit: descriptive names. See above.

src/Okta.Sdk.IntegrationTests/UserTypesClientShould.cs
```diff
@@ -0,0 +1,269 @@
...
+                    Description = $"{nameof(ListAllUserTypes)} Test Description ({i})",
+                    DisplayName = $"{nameof(ListAllUserTypes)} Test DisplayName ({i})",
+                    Name = $"{nameof(ListAllUserTypes)}_TestUserType_{i}",
+                });
+                testUserTypeIds.Add(createdUserType.Id);
+            }
+
+            try
+            {
+                HashSet<string> allUserTypeIds = new HashSet<string>();
```
laura-rodriguez, line 235: nit: In this method we have vars and explicitly typed vars, let's try to use the same coding rules when it's possible for consistency. In general, this SDK is using vars, so I would stick to it.

PR #379, opened 2020-05-19
OKTA-278598: add integration and unit tests for eventhooks

bryanapellanes-okta:
## Issue \#
OKTA-278598: add integration and unit tests for eventhooks

## Code

- [x] Unit test(s)
- [x] Integration tests

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
+++ b/src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
@@ -1,0 +1,382 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Okta.Sdk.IntegrationTests
+{
+    public class EventHooksClientShould
+    {
+        private const string SdkPrefix = "dotnet_sdk";
+        private const string EventType = "EVENT_TYPE";
+
+        private static readonly string[] TestEventItems = new string[]
+        {
+            "user.lifecycle.create",
+            "user.lifecycle.activate",
+        };
+
+        private static readonly string[] UpdatedTestEventItems = new string[]
+        {
+            "user.lifecycle.create",
+            "user.lifecycle.activate",
+            "user.lifecycle.deactivate",
+        };
+
+        private static readonly EventHookChannel TestEventHookChannel = new EventHookChannel
+        {
+            Type = "HTTP",
+            Version = "1.0.0",
+            Config = new EventHookChannelConfig
+            {
+                Uri = "https://www.example.com/eventHooks",
+                Headers = new List<IEventHookChannelConfigHeader>
+                {
+                    new EventHookChannelConfigHeader
+                    {
+                        Key = "X-Test-Header",
+                        Value = "Test header value",
+                    },
+                },
+                AuthScheme = new EventHookChannelConfigAuthScheme
+                {
+                    Type = "HEADER",
+                    Key = "Authorization",
+                    Value = "Test-Api-Key",
+                },
+            },
+        };
+
+        private static readonly EventHookChannel UpdatedTestEventHookChannel = new EventHookChannel
+        {
+            Type = "HTTP",
+            Version = "1.0.0",
+            Config = new EventHookChannelConfig
+            {
+                Uri = "https://www.example.com/eventHooksUpdated",
+                Headers = new List<IEventHookChannelConfigHeader>
+                {
+                    new EventHookChannelConfigHeader
+                    {
+                        Key = "X-Test-Header",
+                        Value = "Test header value updated",
+                    },
+                },
+                AuthScheme = new EventHookChannelConfigAuthScheme
+                {
+                    Type = "HEADER",
+                    Key = "Authorization",
+                    Value = "Test-Api-Key-Updated",
+                },
+            },
+        };
+
+        public EventHooksClientShould()
+        {
+            DeleteAllEventHooks().Wait();
+        }
+
+        [Fact]
+        public async Task CreateEventHook()
+        {
+            var testClient = TestClient.Create();
+            var testEventHookName = $"{SdkPrefix}:{nameof(CreateEventHook)} Test Event Hook Name";
+
+            var createdEventHook = await testClient.EventHooks.CreateEventHookAsync(new EventHook
+            {
+                Name = testEventHookName,
+                Events = new EventSubscriptions
+                {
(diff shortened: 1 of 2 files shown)


src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
```diff
@@ -0,0 +1,382 @@
...
+                AuthScheme = new EventHookChannelConfigAuthScheme
+                {
+                    Type = "HEADER",
+                    Key = "Authorization",
+                    Value = "Test-Api-Key",
+                },
+            },
+        };
+
+        private static readonly EventHookChannel UpdatedTestEventHookChannel = new EventHookChannel
```
laura-rodriguez, line 52: Suggestion: I know we are trying to avoid duplicated code here. However, I think for testing in general, having inline code helps with readability, even if it means duplicated code. I think we should favor code readability here, especially because usually when we add a new API, the reviewer has to take a look at a considerable number of tests, and we should facilitate that process. Let me know your thoughts, I'm open to discussion.

bryanapellanes-okta: You make a good point; moving these definitions closer to where they are used will likely make the tests more easy to read and understand.  Since these definitions aren't referenced in a whole bunch of places I'll move them into the tests that rely on them.

src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
```diff
@@ -0,0 +1,382 @@
...
+                createdEventHook.Events.Should().NotBeNull();
+                createdEventHook.Events.Items.Should().NotBeNull();
+                createdEventHook.Events.Items.Count.Should().Be(TestEventItems.Length);
+                createdEventHook.Channel.Should().NotBeNull();
+                createdEventHook.Channel.Config.Should().NotBeNull();
+                createdEventHook.Channel.Config.Uri.Should().Be(TestEventHookChannel.Config.Uri);
+            }
+            finally
+            {
+                await DeleteAllEventHooks();
```
laura-rodriguez, line 111: Nit: Per [MS guideline](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#avoid-logic-in-tests) we should avoid logical conditions in tests (`foreach` in this case). I recommend removing the specific Event Hook you created on each specific test. That way, if a test fails, we can identify faster what particular entity is failing and analyze the reasons.

src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
```diff
@@ -0,0 +1,382 @@
...
+            DeleteAllEventHooks().Wait();
+        }
+
+        [Fact]
+        public async Task CreateEventHook()
+        {
+            var testClient = TestClient.Create();
+            var testEventHookName = $"{SdkPrefix}:{nameof(CreateEventHook)} Test Event Hook Name";
+
+            var createdEventHook = await testClient.EventHooks.CreateEventHookAsync(new EventHook
```
laura-rodriguez, line 87: Nit: Move `new EventHook` to the next line. [This rule](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1116.md) should be covered by Stylecop. Does the warning show up on VS?

src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
```diff
@@ -0,0 +1,382 @@
...
+            }
+        }
+
+        [Fact]
+        public async Task RetrieveEventHook()
+        {
+            var testClient = TestClient.Create();
+            var testEventHookName = $"{SdkPrefix}:{nameof(RetrieveEventHook)} Test Event Hook Name";
+
+            var createdEventHook = await testClient.EventHooks.CreateEventHookAsync(new EventHook
```
laura-rodriguez, line 121: Check StyleCop comments above.

src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
bryanapellanes-okta: Visual Studio isn't giving me those warnings.  Do we have a stylecop settings file I can import?

laura-rodriguez: I can see that the IT project has StyleCop dependencies and also a stylecop.json file. Warnings should show up on Visual Studio 🤔

src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
```diff
@@ -0,0 +1,382 @@
...
+                    Channel = TestEventHookChannel,
+                });
+                testEventHookIds.Add(createdEventHook.Id);
+            }
+
+            try
+            {
+                var allEventHookIds = new HashSet<string>();
+                var allEventHooks = testClient.EventHooks.ListEventHooks();
+                int allEventHooksCount = await allEventHooks.CountAsync();
```
laura-rodriguez, line 282: Nit: Let's use var here.

src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
bryanapellanes-okta: I'll fix the code and review my stylecop settings separately.

src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
bryanapellanes-okta: Removed DeleteAllEventHooks method in favor of individual deletes per test.

src/Okta.Sdk.IntegrationTests/EventHooksClientShould.cs
```diff
@@ -0,0 +1,524 @@
...
+    public class EventHooksClientShould
+    {
+        private const string SdkPrefix = "dotnet_sdk";
+        private const string EventType = "EVENT_TYPE";
+        
+        [Fact]
+        public async Task CreateEventHook()
+        {
+            var testClient = TestClient.Create();
+            var testEventHookName = $"{SdkPrefix}:{nameof(CreateEventHook)} Test Event Hook Name";
```
laura-rodriguez, line 19: If names have to be unique, then we have to append a random string here.

bryanapellanes-okta: Agreed, I was going to address that concern in a PR for work on OKTA-298370.

PR #382, opened 2020-05-21
Add tests for Admin Roles API

laura-rodriguez:
## Issue \#
OKTA-285603

## Code

- [x] Unit test(s)
- [ ] Implementation

## Additional Context

- Update generated code using the latest version of the openapi spec
- Add ITs and UTs 
- Update the StyleCop header template, so new files use the updated template.

### DIFF
--- a/src/Okta.Sdk/Generated/IUserType.Generated.cs
+++ b/src/Okta.Sdk/Generated/IUserType.Generated.cs
@@ -25,7 +25,7 @@ namespace Okta.Sdk
 
         string DisplayName { get; set; }
 
-        string Id { get; set; }
+        string Id { get; }
 
         DateTimeOffset? LastUpdated { get; }
 
--- a/src/Okta.Sdk/Generated/IUserType.Generated.cs
+++ b/src/Okta.Sdk/Generated/IUserType.Generated.cs
@@ -25,7 +25,7 @@ namespace Okta.Sdk
 
         string DisplayName { get; set; }
 
-        string Id { get; }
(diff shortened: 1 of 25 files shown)


src/Okta.Sdk/Generated/IUserType.Generated.cs
```diff
@@ -25,7 +25,7 @@ public partial interface IUserType : IResource
 
         string DisplayName { get; set; }
 
-        string Id { get; set; }
+        string Id { get; }
```
bryanapellanes-okta, line 28: This might be problematic, I believe this should not be ready only.

src/Okta.Sdk/Generated/UserType.Generated.cs
```diff
@@ -40,11 +40,7 @@ public string DisplayName
         }
         
         /// <inheritdoc/>
-        public string Id 
-        {
-            get => GetStringProperty("id");
-            set => this["id"] = value;
-        }
+        public string Id => GetStringProperty("id");
```
bryanapellanes-okta, line 43: See previous comment; this may cause existing code to not compile.  Did you rebase?

src/Okta.Sdk/Generated/IUserType.Generated.cs
laura-rodriguez: Ok, good to know. It changed after I ran the generation process using the latest version. I had the same issue with other entities. I will fix it in the spec.

PR #383, opened 2020-05-21
OKTA-285614 add linked objects

bryanapellanes-okta:
## Issue \#
OKTA-285614

## Code
- [x] Integration test(s)
- [ ] Implementation

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/DisableTestParallelization.cs
+++ b/src/Okta.Sdk.IntegrationTests/DisableTestParallelization.cs
@@ -1,0 +1,3 @@
+﻿using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
--- a/src/Okta.Sdk.IntegrationTests/DisableTestParallelization.cs
+++ b/src/Okta.Sdk.IntegrationTests/DisableTestParallelization.cs
@@ -1,3 +1,8 @@
-﻿using Xunit;
+﻿// <copyright file="DisableTestParallelization.cs" company="Okta, Inc">
+// Copyright (c) 2020 - present Okta, Inc. All rights reserved.
+// Licensed under the Apache 2.0 license. See the LICENSE file in the project root for full license information.
+// </copyright>
+
+using Xunit;
 
 [assembly: CollectionBehavior(DisableTestParallelization = true)]
--- a/src/Okta.Sdk.IntegrationTests/OAuthScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/OAuthScenarios.cs
@@ -16,7 +16,7 @@ namespace Okta.Sdk.IntegrationTests
 {
     public class OAuthScenarios
     {
-        [Fact]
+        [Fact(Skip = "Investigate collateral errors (404 - CVDType) - OKTA-300462")]
         [Trait("Category", "NoBacon")] // Tests that don't run on internal CI pipeline
         public async Task ListUsers()
         {
--- a/src/Okta.Sdk.IntegrationTests/OAuthScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/OAuthScenarios.cs
@@ -16,7 +16,7 @@ namespace Okta.Sdk.IntegrationTests
 {
     public class OAuthScenarios
     {
-        [Fact(Skip = "Investigate collateral errors (404 - CVDType) - OKTA-300462")]
+        [Fact]
         [Trait("Category", "NoBacon")] // Tests that don't run on internal CI pipeline
         public async Task ListUsers()
         {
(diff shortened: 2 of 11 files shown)


regenerate
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+
```
laura-rodriguez, line 3: Since we are not gonna use gitmodules at the moment due to openapi private repo, let's exclude related files.

src/Okta.Sdk.IntegrationTests/LinkedObjectClientShould.cs
```diff
@@ -0,0 +1,323 @@
+﻿using System.Collections.Generic;
```
laura-rodriguez, line 1: Make sure to include StyleCop header. VS should hint you at the top of the file to include it.
Also, to be consistent with our other ITs let's rename this file/class to LinkedObjectScenarios. I know we have mixed names, but the rule is if the IT we are creating is following an API reference (Applications, Users, Groups, etc) the name should be <API>Scenarios, otherwise <X>Should.

src/Okta.Sdk.IntegrationTests/DisableTestParallelization.cs
```diff
@@ -0,0 +1,3 @@
+﻿using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
laura-rodriguez, line 3: question: Do you have to create a new class for this setting to work? I wonder if we can configure this via xunit json or something like that?

bryanapellanes-okta: The information I read implied that xunit is scanning assemblies to determine appropriate behavior.  This setting is technically not in a class it's sitting in the global/unnamed namespace.

bryanapellanes-okta: Something like this would typically be placed in the AssemblyInfo.cs file, but because dotnet core will generate that file depending on compile time settings, putting it in it's own file seemed logical to me.

laura-rodriguez: It seems you can still configure this via csproject or xunit.runner.json. Have you tried this? See post #4 here: https://stackoverflow.com/questions/1408175/execute-unit-tests-serially-rather-than-in-parallel. If configuring via csproject or xunit.runner.json is possible, I think we should go that way. I feel it is more intuitive to have this type of configuration in some sort of configuration file.

bryanapellanes-okta: Certainly something we can try.  I'd like to reserve that for a future experiment considering the challenges faced getting this build green.

src/Okta.Sdk.IntegrationTests/UserTypesClientShould.cs
```diff
@@ -168,7 +168,6 @@ public async Task ReplaceUserType()
             }
         }
```
laura-rodriguez, line 169: We should rename this file to follow the "scenarios" pattern. You can do this separately if you want.

src/Okta.Sdk.UnitTests/LinkedObjectClientShould.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
```
laura-rodriguez, line 1: Missing copyright header?

bryanapellanes-okta: The header is in the follow up changes.

src/Okta.Sdk.IntegrationTests/OAuthScenarios.cs
```diff
@@ -16,7 +16,7 @@ namespace Okta.Sdk.IntegrationTests
 {
     public class OAuthScenarios
     {
-        [Fact]
+        [Fact(Skip = "Investigate collateral errors (404 - CVDType) - OKTA-300462")]
```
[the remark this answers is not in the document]
bryanapellanes-okta: Yes, there is an open core issue related to anything that creates users.  I'm addressing some of these "Skips" in OKTA-300462.

src/Okta.Sdk.UnitTests/LinkedObjectClientShould.cs
```diff
@@ -0,0 +1,35 @@
...
+using Xunit;
+
+namespace Okta.Sdk.UnitTests
+{
+    public class LinkedObjectClientShould
+    {
+        private const string SdkPrefix = "dotnet_sdk";
+
+        [Fact]
+        public async Task CallDeleteEndpoint()
```
laura-rodriguez, line 18: Can `DeleteLinkedObjectDefinitionAsync` be tested via IT instead?

bryanapellanes-okta: Due to server race conditions I was planning on doing the integration test for this after addressing some of the outstanding concerns.

src/Okta.Sdk.UnitTests/LinkedObjectClientShould.cs
bryanapellanes-okta: Correction; I'll update.

PR #386, opened 2020-05-29
OKTA-285610: add tests for features client

bryanapellanes-okta:
## Code

- [x] Unit test(s)
- [x] Integration test(s)

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
@@ -26,7 +26,7 @@ namespace Okta.Sdk.IntegrationTests
         public async Task GetFeatureById()
         {
             var client = TestClient.Create();
-            var feature = (await client.Features.ListFeatures().ToListAsync()).FirstOrDefault();
+            var feature = await client.Features.ListFeatures().FirstOrDefaultAsync();
             feature.Should().NotBeNull();
 
             var retrievedFeature = await client.Features.GetFeatureAsync(feature.Id);
--- a/src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
@@ -1,0 +1,61 @@
+﻿// <copyright file="FeaturesScenarios.cs" company="Okta, Inc">
+// Copyright (c) 2020 - present Okta, Inc. All rights reserved.
+// Licensed under the Apache 2.0 license. See the LICENSE file in the project root for full license information.
+// </copyright>
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Xunit;
+
+namespace Okta.Sdk.IntegrationTests
+{
+    public class FeaturesScenarios
+    {
+        [Fact]
+        public async Task ListFeatures()
+        {
+            var client = TestClient.Create();
+            var features = await client.Features.ListFeatures().ToListAsync();
+            features.Should().NotBeNull();
+            features.Count.Should().BeGreaterThan(0);
+        }
+
+        [Fact]
+        public async Task GetFeatureById()
+        {
+            var client = TestClient.Create();
+            var feature = (await client.Features.ListFeatures().ToListAsync()).FirstOrDefault();
+            feature.Should().NotBeNull();
+
+            var retrievedFeature = await client.Features.GetFeatureAsync(feature.Id);
+            retrievedFeature.Should().NotBeNull();
+            retrievedFeature.Id.Should().Be(feature.Id);
+            retrievedFeature.Name.Should().Be(feature.Name);
+            retrievedFeature.Description.Should().Be(feature.Description);
+        }
+
+        [Fact]
+        public async Task GetDependencies()
+        {
+            var client = TestClient.Create();
+            var knownFeatureId = "ftrygtN0AtAqTxM5B4X9"; // Allow end-users to quickly access everyday apps in the Okta Browser Plugin
+            var dependencies = await client.Features.ListFeatureDependencies(knownFeatureId).ToListAsync();
+
+            dependencies.Should().NotBeNull();
+            dependencies.Count.Should().BeGreaterThan(0);
+        }
+
+        [Fact]
+        public async Task GetDependents()
+        {
+            var client = TestClient.Create();
+            var knownFeatureId = "ftrFDVSQ9wBWPnDqohOS"; // Allow end-users to quickly access everyday apps
+            var dependents = await client.Features.ListFeatureDependents(knownFeatureId).ToListAsync();
+
+            dependents.Should().NotBeNull();
+            dependents.Count.Should().BeGreaterThan(0);
+        }
+    }
+}
--- a/src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
@@ -3,7 +3,6 @@
 // Licensed under the Apache 2.0 license. See the LICENSE file in the project root for full license information.
 // </copyright>
 
-using System;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


src/Okta.Sdk.UnitTests/FeaturesClientShould.cs
```diff
@@ -0,0 +1,36 @@
...
+using Okta.Sdk.Internal;
+using Okta.Sdk.UnitTests.Internal;
+using Xunit;
+
+namespace Okta.Sdk.UnitTests
+{
+    public class FeaturesClientShould
+    {
+        [Fact]
+        public async Task CallUpdateFeatureLifeCycleEndpoint()
```
bryanapellanes-okta, line 19: Due to concerns with tests that modify data (considering potential for server related issues), I opted to implement this test as a unit test using mocks instead of a more detailed integration test.  We can certainly revisit the creation of integration tests for this as time allows.

src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Okta.Sdk.IntegrationTests
+{
+    public class FeaturesScenarios
+    {
+        [Fact]
+        public async Task ListFeatures()
+        {
+            var client = TestClient.Create();
+            var features = await client.Features.ListFeatures().ToListAsync();
```
laura-rodriguez, line 20: I think you have to create at least a feature here before listing, otherwise, we are trusting in something we don't have control.

src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
```diff
@@ -0,0 +1,61 @@
...
+            features.Should().NotBeNull();
+            features.Count.Should().BeGreaterThan(0);
+        }
+
+        [Fact]
+        public async Task GetFeatureById()
+        {
+            var client = TestClient.Create();
+            var feature = (await client.Features.ListFeatures().ToListAsync()).FirstOrDefault();
```
laura-rodriguez, line 29: You can also use `await FirstOrDefaultAsync` to avoid those parentheses ;).

src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
```diff
@@ -0,0 +1,61 @@
...
+            feature.Should().NotBeNull();
+
+            var retrievedFeature = await client.Features.GetFeatureAsync(feature.Id);
```
laura-rodriguez, line 32: Is it possible to create a feature and test with it?

src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
bryanapellanes-okta: Generally I would agree; however in my own experimentation/testing, it appears that features themselves are integral to the system, without which we wouldn't have sets of valid offerings.  Also, I try to avoid making data updates that may be difficult to clean up later.

src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
bryanapellanes-okta: I don't believe it would be a trivial thing to create a feature for the purposes of testing, there is no api exposed for this purpose.  Features themselves are the units of functionality that we offer to customers and the list of features is a well know/defined list.  It would potentially cause broader issues if we added features purely for testing.

src/Okta.Sdk.UnitTests/FeaturesClientShould.cs
laura-rodriguez: I know we have been struggling with lots of IT issues. However, our default should still be creating ITs unless it is too complicated or we are actually seeing issues that are difficult to solve. The idea is that we help to identify problems with our API.

src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
bryanapellanes-okta: Good call!

src/Okta.Sdk.UnitTests/FeaturesClientShould.cs
bryanapellanes-okta: I review further.

src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
laura-rodriguez: Ah! I see. So, if there's no API provided let's confirm that its exclusion was intended (I think in this case makes sense).

src/Okta.Sdk.UnitTests/FeaturesClientShould.cs
laura-rodriguez: Now that I know that there's no API provided to create our own feature and play with it, I think you made a good decision on writing a UT for this case instead 👍 .

src/Okta.Sdk.IntegrationTests/FeaturesScenarios.cs
```diff
@@ -0,0 +1,61 @@
...
+            retrievedFeature.Id.Should().Be(feature.Id);
+            retrievedFeature.Name.Should().Be(feature.Name);
+            retrievedFeature.Description.Should().Be(feature.Description);
+        }
+
+        [Fact]
+        public async Task GetDependencies()
+        {
+            var client = TestClient.Create();
+            var knownFeatureId = "ftrygtN0AtAqTxM5B4X9"; // Allow end-users to quickly access everyday apps in the Okta Browser Plugin
```
laura-rodriguez, line 43: Given that we are not certain that dependencies and dependents relationships will stay the same in the future, I would write UTs for `GetDependencies` and `GetDependents` instead.

bryanapellanes-okta: will do!

src/Okta.Sdk.UnitTests/FeaturesClientShould.cs
```diff
@@ -0,0 +1,81 @@
...
+            mockResponse.Payload =
+                "[{\"Description\":null,\"Id\":null,\"Name\":\"TestFeature\",\"Stage\":{\"State\":null,\"Value\":null},\"Status\":null,\"Type\":null}]";
+            mockResponse.StatusCode = 200;
+            _ = mockRequestExecutor
+                .GetAsync(dependenciesEndpoint, Arg.Any<Dictionary<string, string>>(), Arg.Any<CancellationToken>())
+                .Returns(Task.FromResult(mockResponse));
+
+            var testClient = new TestableOktaClient(mockRequestExecutor);
+
+            _ = await testClient.Features.ListFeatureDependencies(featureId).ToListAsync();
```
laura-rodriguez, line 54: When testing a list/get, we could also test that the payload you sent is translated into a valid object in your list. For example:
```
rertievedDependencies.FirstOrDefault().Name.Should().Be("TestFeature");

File: src/Okta.Sdk.UnitTests/FeaturesClientShould.cs [hunk 7]
```diff
@@ -0,0 +1,81 @@
...
+            mockResponse.Payload =
+                "[{\"Description\":null,\"Id\":null,\"Name\":\"TestFeature\",\"Stage\":{\"State\":null,\"Value\":null},\"Status\":null,\"Type\":null}]";
+            mockResponse.StatusCode = 200;
+            _ = mockRequestExecutor
+                .GetAsync(dependentsEndpoint, Arg.Any<Dictionary<string, string>>(), Arg.Any<CancellationToken>())
+                .Returns(Task.FromResult(mockResponse));
+
+            var testClient = new TestableOktaClient(mockRequestExecutor);
+
+            _ = await testClient.Features.ListFeatureDependents(featureId).ToListAsync();
```
laura-rodriguez, line 76: Same above comment here.

PR #387, opened 2020-05-29
OKTA-300462 account for user race condition

bryanapellanes-okta:
# Account for eventual consistency

The primary change introduced here are Thread.Sleep additions on creation of new data.  Was seeing intermittent 404 responses when new data entries are made, likely due to GET requests going to read only data replicas.  This change waits a few seconds to allow for data replication prior to read attempts.

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/EventHooksClientScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/EventHooksClientScenarios.cs
@@ -12,7 +12,7 @@ using Xunit.Sdk;
 
 namespace Okta.Sdk.IntegrationTests
 {
-    public class EventHooksClientShould
+    public class EventHooksClientScenarios
     {
         private const string SdkPrefix = "dotnet_sdk";
         private const string EventType = "EVENT_TYPE";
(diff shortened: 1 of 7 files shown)


src/Okta.Sdk.IntegrationTests/EventHooksClientScenarios.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace Okta.Sdk.IntegrationTests
 {
-    public class EventHooksClientShould
+    public class EventHooksClientScenarios
```
laura-rodriguez, line 15: Make sure to update the header as well.

src/Okta.Sdk.IntegrationTests/InlineHooksClientScenarios.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Okta.Sdk.IntegrationTests
 {
-    public class InlineHooksClientShould
+    public class InlineHooksClientScenarios
```
laura-rodriguez, line 14: Make sure to update the header as well.

src/Okta.Sdk.IntegrationTests/UserTypesClientScenarios.cs
```diff
@@ -6,13 +6,14 @@
...
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Xunit;
 
 namespace Okta.Sdk.IntegrationTests
 {
-    public class UserTypesClientShould
+    public class UserTypesClientScenarios
```
laura-rodriguez, line 16: Make sure to update the header as well.

PR #392, opened 2020-06-17
OKTA-295020 support for different content types

bryanapellanes-okta:
## Issue \#
OKTA-295020
OKTA-302822

## Code

- [x] Unit test(s)
- [x] Integration test(s)
- [x] Implementation

## Current behavior
The OktaClient is hard coded to only send Content-Type headers with the value of "application/json".

## Desired behavior
Enable a developer to override the default Content-Type header with an alternative value when necessary.

## Additional Context
Some endpoints require Content-Type header values other than "application/json", see https://developer.okta.com/docs/reference/api/apps/#request-example-45

This PR also includes POST support for collections.


src/Okta.Sdk.IntegrationTests/ApplicationScenarios.cs
```diff
@@ -8,14 +8,6 @@
...
 using System.Threading.Tasks;
 using FluentAssertions;
-using Org.BouncyCastle.Asn1.X509;
-using Org.BouncyCastle.Crypto;
-using Org.BouncyCastle.Crypto.Generators;
-using Org.BouncyCastle.Crypto.Prng;
-using Org.BouncyCastle.Math;
-using Org.BouncyCastle.Security;
-using Org.BouncyCastle.Utilities;
-using Org.BouncyCastle.X509;
```
laura-rodriguez, line 18: Since we are not planning to use this library, let's remove the NuGet dependency as well.

src/Okta.Sdk.UnitTests/DefaultRequestExecutorShould.cs
```diff
@@ -85,5 +85,75 @@ public async Task NotRetryRequestOnceWhenResponseIsNot401AndAuthorizationModeIsP
...
         }
+
+        [Theory]
+        [InlineData(HttpVerb.Get)]
+        [InlineData(HttpVerb.Post)]
+        [InlineData(HttpVerb.Put)]
+        [InlineData(HttpVerb.Delete)]
+        public async Task ExecuteMethodBasedOnVerb(HttpVerb verb)
+        {
+            var requestMessageHandler = new MockHttpMessageHandler(string.Empty, HttpStatusCode.OK);
```
laura-rodriguez, line 96: I think we can simplify this test by removing OAuth related code unless you specifically want to test expected verbs when using OAuth.

src/Okta.Sdk.UnitTests/DefaultRequestExecutorShould.cs
```diff
@@ -85,5 +85,75 @@ public async Task NotRetryRequestOnceWhenResponseIsNot401AndAuthorizationModeIsP
...
+            testRequestExecutor.VerbExecutionCounts[verb].Should().Be(1);
+        }
+
+        [Fact]
+        public async Task SetMessageContentForPost()
+        {
+            var requestMessageHandler = new MockHttpMessageHandler(string.Empty, HttpStatusCode.OK);
+            var httpClientRequest = new HttpClient(requestMessageHandler);
+            var configuration = new OktaClientConfiguration
+            {
```
laura-rodriguez, line 131: Same comment here, do we need OAuth code related here? Can we simplify this test?

src/Okta.Sdk.UnitTests/FeaturesClientShould.cs
```diff
@@ -47,7 +46,7 @@ public async Task CallDependenciesEndpoint()
                 "[{\"Description\":null,\"Id\":null,\"name\":\"TestFeature\",\"Stage\":{\"State\":null,\"Value\":null},\"Status\":null,\"Type\":null}]";
             mockResponse.StatusCode = 200;
             _ = mockRequestExecutor
-                .GetAsync(dependenciesEndpoint, Arg.Any<Dictionary<string, string>>(), Arg.Any<CancellationToken>())
+                .ExecuteRequestAsync(Arg.Any<HttpRequest>(), Arg.Any<CancellationToken>())
```
laura-rodriguez, line 49: I could see you still use `PostAsync` above, but you are replacing `GetAsync` by `ExecuteRequestAsync` here, Is `GetAsync` no longer available?

src/Okta.Sdk.UnitTests/FeaturesClientShould.cs
```diff
@@ -70,15 +69,16 @@ public async Task GetDependents()
...
                 .Returns(Task.FromResult(mockResponse));
 
             var testClient = new TestableOktaClient(mockRequestExecutor);
 
             var retrievedDependents = await testClient.Features.ListFeatureDependents(featureId).ToListAsync();
             retrievedDependents.FirstOrDefault().Name.Should().Be("TestFeature");
             _ = await mockRequestExecutor.Received(1)
-                .GetAsync(dependentsEndpoint, Arg.Any<Dictionary<string, string>>(), Arg.Any<CancellationToken>());
+                .Received(1)
```
laura-rodriguez, line 80: It seems we have `.Received(1)` twice, do we want that?

src/Okta.Sdk.UnitTests/Internal/MockedStringRequestExecutor.cs
```diff
@@ -5,13 +5,14 @@
...
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Okta.Sdk.Internal;
 
 namespace Okta.Sdk.UnitTests.Internal
 {
-    public class MockedStringRequestExecutor : IRequestExecutor
+    public class MockedStringRequestExecutor : DefaultRequestExecutor
     {
```
laura-rodriguez, line 16: Shouldn't we implement the interface instead?

src/Okta.Sdk.UnitTests/Internal/TestDefaultRequestExecutor.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Okta.Sdk.Configuration;
+using Okta.Sdk.Internal;
+
+namespace Okta.Sdk.UnitTests.Internal
+{
+    public class TestDefaultRequestExecutor : DefaultRequestExecutor
```
laura-rodriguez, line 11: Shouldn't we implement the interface instead?

src/Okta.Sdk/HttpRequest.cs
```diff
@@ -29,6 +53,21 @@ public class HttpRequest
         /// </value>
         public object Payload { get; set; }
 
+        private string _contentType;
```
laura-rodriguez, line 56: Nit: Let's place private attributes at the top of the file.

src/Okta.Sdk/HttpRequest.cs
```diff
@@ -29,6 +53,21 @@ public class HttpRequest
...
+        /// <summary>
+        /// Gets or sets the content type.
+        /// </summary>
+        public string ContentType
+        {
+            get => _contentType;
+            set
+            {
+                _contentType = value;
+                PayloadHandler = Internal.PayloadHandler.ForContentType(_contentType);
```
laura-rodriguez, line 67: Nit: Let's remove Internal from here and add the corresponding namespace.

src/Okta.Sdk/HttpRequest.cs
```diff
@@ -55,5 +94,39 @@ public class HttpRequest
         /// </value>
         public IDictionary<string, string> Headers { get; set; }
             = new Dictionary<string, string>();
+
+        private string _body;
```
laura-rodriguez, line 98: NIt: PEr convention, let's move private attributes at the top of the file.

src/Okta.Sdk/IOktaClient.cs
```diff
@@ -16,6 +14,25 @@ namespace Okta.Sdk
     /// </summary>
     public interface IOktaClient
     {
+        /// <summary>
+        /// Gets the request context.
+        /// </summary>
+        RequestContext RequestContext { get; }
```
laura-rodriguez, line 20: Do we need to expose this?

src/Okta.Sdk/Internal/DefaultDataStore.cs
```diff
@@ -140,6 +140,11 @@ private static void ApplyContextToRequest(HttpRequest request, RequestContext co
             {
                 request.Headers["Content-Transfer-Encoding"] = context.ContentTransferEncoding;
             }
+
+            if (!string.IsNullOrEmpty(context.ContentType))
+            {
+                request.ContentType = context.ContentType;
```
laura-rodriguez, line 146: What happens if contentType is null or empty? Do we have a default?

src/Okta.Sdk/Internal/DefaultDataStore.cs
```diff
@@ -222,8 +227,8 @@ public async Task<HttpResponse<TResponse>> PostAsync<TResponse>(HttpRequest requ
             PrepareRequest(request, requestContext);
 
             var body = _serializer.Serialize(request.Payload);
-
-            var response = await _requestExecutor.PostAsync(request.Uri, request.Headers, body, cancellationToken).ConfigureAwait(false);
+            request.SetBody(body);
```
laura-rodriguez, line 230: Since we have a `PreparaRequest` method, should we move this code there?

src/Okta.Sdk/Internal/DefaultRequestExecutor.cs
```diff
@@ -216,5 +254,18 @@ public Task<HttpResponse<string>> DeleteAsync(string href, IEnumerable<KeyValueP
 
             return SendAsync(request, cancellationToken);
         }
+
+        private HttpRequestMessage CreateMessage(HttpRequest httpRequest)
```
laura-rodriguez, line 258: Nit: I would rename this method to be more specific, something like `CreateHttpRequestMessage`.

src/Okta.Sdk/Internal/DefaultRequestExecutor.cs
```diff
@@ -216,5 +254,18 @@ public Task<HttpResponse<string>> DeleteAsync(string href, IEnumerable<KeyValueP
...
+        {
+            var path = EnsureRelativeUrl(httpRequest.Uri);
+            var httpRequestMessage = new HttpRequestMessage(_httpMethods[httpRequest.Verb], new Uri(path, UriKind.Relative));
+            ApplyHeadersToRequest(
+                httpRequestMessage,
+                httpRequest.Headers.Keys
+                    .Select(header => new KeyValuePair<string, string>(header, httpRequest.Headers[header]))
+                    .ToList());
+            httpRequest.SetMessageContent(httpRequestMessage);
```
laura-rodriguez, line 267: Nit: Same here. Let's rename this method too to be more specific.

src/Okta.Sdk/Internal/PayloadHandler.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        /// <inheritdoc/>
+        public string ContentType { get; protected set; }
+
+        /// <summary>
+        /// Gets or sets the content transfer encoding.
+        /// </summary>
+        protected string ContentTransferEncoding { get; set; }
+
+        private string _body;
```
laura-rodriguez, line 44: Nit: Let's place private attributes at the top of the file.

src/Okta.Sdk/Internal/PayloadHandler.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PayloadHandler"/> class.
+        /// </summary>
+        public PayloadHandler()
+        {
+            Serializer = new DefaultSerializer();
+        }
+
+        static PayloadHandler()
```
laura-rodriguez, line 27: Missing access modifier?

src/Okta.Sdk/Internal/PayloadHandler.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            return _body;
+        }
+
+        /// <summary>
+        /// Get content for the specified request.
+        /// </summary>
+        /// <param name="httpRequest">The request whose content is returned.</param>
+        /// <returns>Content for the specified request.</returns>
+        protected abstract HttpContent GetContent(HttpRequest httpRequest);
```
laura-rodriguez, line 66: Nit: Let's rename this method to be more specific, something like: `GetRequestHttpContent`

src/Okta.Sdk.UnitTests/DefaultRequestExecutorShould.cs
bryanapellanes-okta: I'll see about removing those bits.  I took a lot of the setup code from other tests and wasn't sure which parts were specifically needed for the mocks to function correctly.

src/Okta.Sdk.UnitTests/FeaturesClientShould.cs
bryanapellanes-okta: ExecuteRequestAsync will delegate to the appropriate method (PostAsync, GetAsync, etc.) based on the verb specified in the HttpRequest.  There are different layers delegating downward, this change was necessary for the logic of this test to represent the new underlying behavior.

src/Okta.Sdk.UnitTests/FeaturesClientShould.cs
bryanapellanes-okta: Good catch, I'll fix.

src/Okta.Sdk.UnitTests/Internal/MockedStringRequestExecutor.cs
bryanapellanes-okta: The DefaultRequestExecutor is the primary implementation and has internals that are hard to mock correctly.  I would tend to agree, however, the MockedStringRequestExecutor is used to gain access to state and behavior for validation purposes.  I think a more appropriate change would be to rename MockedStringRequestExecutor to TestStringRequestExecutor but that feels more like semantics.

src/Okta.Sdk.UnitTests/Internal/TestDefaultRequestExecutor.cs
bryanapellanes-okta: See previous comment.

src/Okta.Sdk/HttpRequest.cs
bryanapellanes-okta: I can do that if we prefer, however my Visual Studio settings advise against it:
image

Is there a stylecop setting that I'm missing?

src/Okta.Sdk/IOktaClient.cs
bryanapellanes-okta: It seemed useful to do so; I'll see if not doing so is feasible.

src/Okta.Sdk/Internal/DefaultDataStore.cs
bryanapellanes-okta: The default content type is always "application/json".  This logic determines if that default should be overridden by what's specified in the request context.

src/Okta.Sdk/Internal/DefaultDataStore.cs
bryanapellanes-okta: Setting the body in this way is only appropriate here due to the way that the body/payload is handled in the broader context.

src/Okta.Sdk/Internal/PayloadHandler.cs
bryanapellanes-okta: Default is private, I'll add it in explicitly.

src/Okta.Sdk.UnitTests/DefaultRequestExecutorShould.cs
bryanapellanes-okta: After additional testing, I've found that the existing implementation of NullOAuthTokenProvider throws an exception that is difficult to test around without this additional setup.

src/Okta.Sdk/IOktaClient.cs
bryanapellanes-okta: Removed RequestContext property from IOktaClient.

src/Okta.Sdk.UnitTests/ApplicationsClientShould.cs
```diff
@@ -130,6 +131,61 @@ public async Task BuildRevokeTokensRequest()
...
+
+            await client.Applications.PublishBinaryDerCertAsync(Encoding.UTF8.GetBytes("foo"), "bar", "baz");
+
+            mockRequestExecutor.ReceivedHref.Should().Be("/api/v1/apps/bar/credentials/csrs/baz/lifecycle/publish");
+        }
+
+        [Fact]
+        public async Task BuildPublishBinaryPemCertRequest()
+        {
+            var mockRequestExecutor = new MockedStringRequestExecutor(string.Empty);
```
laura-rodriguez, line 181: For PublishCsr* methods, let's verify that headers and encoding are being sent correctly.

src/Okta.Sdk.UnitTests/DefaultDataStoreShould.cs
```diff
@@ -378,12 +376,34 @@ public async Task AddContextXForwardedToRequests()
...
+
+            var request = new TestHttpRequest();
+            request.ContentType.Should().Be("application/json");
+            request.GetPayLoadHandler().ContentType.Should().Be("application/json");
+
+            var dataStore = new DefaultDataStore(mockRequestExecutor, new DefaultSerializer(), new ResourceFactory(null, null), NullLogger.Instance);
+            await dataStore.PostAsync<TestResource>(request, new RequestContext { ContentType = "foo" }, CancellationToken.None);
+
+            request.ContentType.Should().Be("foo");
+            request.GetPayLoadHandler().ContentType.Should().Be("foo");
```
laura-rodriguez, line 398: Nit: `GetPayloadHandler`

src/Okta.Sdk.UnitTests/DefaultRequestExecutorShould.cs
```diff
@@ -85,5 +85,66 @@ public async Task NotRetryRequestOnceWhenResponseIsNot401AndAuthorizationModeIsP
...
+        [InlineData(HttpVerb.Post)]
+        [InlineData(HttpVerb.Put)]
+        [InlineData(HttpVerb.Delete)]
+        public async Task ExecuteMethodBasedOnVerb(HttpVerb verb)
+        {
+            var httpClientRequest = new HttpClient();
+            var configuration = new OktaClientConfiguration
+            {
+                OktaDomain = "https://myOktaDomain.oktapreview.com",
+                AuthorizationMode = AuthorizationMode.PrivateKey,
```
laura-rodriguez, line 100: Define `Token = "foo"` so you can remove all oAuth code related and simplify this test.

src/Okta.Sdk.UnitTests/DefaultRequestExecutorShould.cs
```diff
@@ -85,5 +85,66 @@ public async Task NotRetryRequestOnceWhenResponseIsNot401AndAuthorizationModeIsP
...
+            var tokenProvider = new DefaultOAuthTokenProvider(configuration, resourceFactory, httpClientToken);
+
+            var testRequestExecutor = new DefaultRequestExecutor(configuration, httpClientRequest, logger, new NoRetryStrategy(), tokenProvider);
+
+            var testRequest = new TestHttpRequest { Uri = "/foo", Verb = HttpVerb.Post };
+            testRequest.HttpRequestMessage.Should().BeNull();
+
+            await testRequestExecutor.PostAsync(testRequest, CancellationToken.None);
+
+            testRequest.HttpRequestMessage.Should().NotBeNull();
```
laura-rodriguez, line 131: Nit: Rename `HttpRequestMessage` to something like `SentHttpRequestMessage`.

src/Okta.Sdk.UnitTests/Internal/MockedCollectionRequestExecutor{T}.cs
```diff
@@ -13,7 +13,7 @@
 
 namespace Okta.Sdk.UnitTests.Internal
 {
-    public class MockedCollectionRequestExecutor<T> : IRequestExecutor
+    public class MockedCollectionRequestExecutor<T> : DefaultRequestExecutor
```
laura-rodriguez, line 16: As discussed, let's implement the interface here and use `DefaultRequestExecutor` as an internal property.

src/Okta.Sdk.UnitTests/OktaClientShould.cs
```diff
@@ -9,13 +9,64 @@
...
+        [Fact]
+        public async Task OverrideRequestContentType()
+        {
+            var executor = Substitute.For<IRequestExecutor>();
+            executor
+                .PostAsync(Arg.Any<HttpRequest>(), Arg.Any<CancellationToken>())
+                .Returns(Task.FromResult(new HttpResponse<string>() { StatusCode = 200 }));
+
+            var client = new TestableOktaClient(executor);
+            await client.CreatedScoped(new RequestContext { ContentType = "foo" }).PostAsync(new HttpRequest { ContentType = "bar" });
```
laura-rodriguez, line 49: Let's mention in the README that `RequestContext.ContentType` > `HttpRequest.ContentType`

src/Okta.Sdk/Internal/DefaultRequestExecutor.cs
```diff
@@ -3,21 +3,21 @@
...
 using System.Threading.Tasks;
-using Microsoft.Extensions.Logging;
-using Okta.Sdk.Configuration;
 
 namespace Okta.Sdk.Internal
 {
     /// <summary>
     /// The default implementation of <see cref="IRequestExecutor"/> that uses <c>System.Net.Http</c>.
     /// </summary>
-    public sealed class DefaultRequestExecutor : IRequestExecutor
```
laura-rodriguez, line 20: Let's keep this.

src/Okta.Sdk/OktaClient.cs
```diff
@@ -69,6 +69,8 @@ public OktaClient(OktaClientConfiguration apiClientConfiguration = null, ILogger
                 serializer,
                 resourceFactory,
                 logger);
+
+            PayloadHandler.TryRegister<PkixCertPayloadHandler>();
```
laura-rodriguez, line 73: Is Json payload already registered?

src/Okta.Sdk/OktaClient.cs
```diff
@@ -144,6 +146,9 @@ private static OktaClientConfiguration GetConfigurationOrDefault(OktaClientConfi
             return compiledConfig;
         }
 
+        /// <inheritdoc/>
+        public RequestContext RequestContext => _requestContext;
```
laura-rodriguez, line 150: As discussed, let's try to avoid exposing this or ultimately making sure is read-only.

src/Okta.Sdk/OktaClient.cs
bryanapellanes-okta: Yes, Json payload is the default and is always registered.

src/Okta.Sdk.UnitTests/ApplicationsClientShould.cs
bryanapellanes-okta: Per our conversation, this won't be possible here due to the nature of the underlying implementation.  Validation for the actual headers occurs in other tests.

src/Okta.Sdk.UnitTests/Internal/MockedCollectionRequestExecutor{T}.cs
bryanapellanes-okta: Per further discussion, the final implementation of this test won't be able to utilize the specific approach you mention; I did change this back to implementing the interface vs extending the base class.

src/Okta.Sdk.UnitTests/OktaClientShould.cs
bryanapellanes-okta: Updated readme: https://github.com/okta/okta-sdk-dotnet/pull/392/commits/258e61c8b7fa765a206cea8728d9e2fa06856b9f

bryanapellanes-okta:
[x] assert that instantiating OktaClient registers PkixCertPayloadHandler (https://github.com/okta/okta-sdk-dotnet/blob/OKTA-295020-content-types/src/Okta.Sdk.UnitTests/OktaClientShould.cs#L347)
[x] assert that setting ContentType sets the PayloadHandler (https://github.com/okta/okta-sdk-dotnet/blob/OKTA-295020-content-types/src/Okta.Sdk.UnitTests/HttpRequestShould.cs#L19)
[ ] assert that Publish* methods call PostAsync<JsonWebKey>(HttpRequest request, Arg.Any<CancellationToken>());  no easy way to do this because ApplicationsClient is sealed (see code: https://github.com/okta/okta-sdk-dotnet/blob/OKTA-295020-content-types/src/Okta.Sdk/ApplicationsClient.cs#L62)
[x] assert that PostAsync calls _dataStore.PostAsync(HttpRequest request, Arg.Any<RequestContext>(), Arg.Any<CancellationToken>()) (https://github.com/okta/okta-sdk-dotnet/blob/OKTA-295020-content-types/src/Okta.Sdk.UnitTests/ApplicationsClientShould.cs#L194)
[x] assert that DefaultDataStore.PostAsync calls _requestExecutor.PostAsync(HttpRequest request, Arg.Any<CancellationToken>()) (https://github.com/okta/okta-sdk-dotnet/blob/OKTA-295020-content-types/src/Okta.Sdk.UnitTests/DefaultDataStoreShould.cs#L402)
[x] assert that DefaultRequestExecutor.PostAsync calls _httpRequestMessageProvider.CreateHttpRequestMessage(HttpRequest request, Arg.Any<string>(), Arg.Any<CancellationToken>() (https://github.com/okta/okta-sdk-dotnet/blob/OKTA-295020-content-types/src/Okta.Sdk.UnitTests/DefaultRequestExecutorShould.cs#L115)
[x] assert that HttpRequestMessageProvider.CreateHttpRequestMessage(HttpRequest request, Arg.Any<string>()) calls HttpRequest.SetHttpRequestMessageContent (https://github.com/okta/okta-sdk-dotnet/blob/OKTA-295020-content-types/src/Okta.Sdk.UnitTests/HttpRequestMessageProviderShould.cs#L16)
[x] assert that HttpRequest.SetHttpRequestMessageContent calls PayloadHandler.SetHttpRequestMessageContent (https://github.com/okta/okta-sdk-dotnet/blob/OKTA-295020-content-types/src/Okta.Sdk.UnitTests/HttpRequestShould.cs#L42)
[x] assert that PayloadHandler.SetHttpRequestMessageContent sets the Content (https://github.com/okta/okta-sdk-dotnet/blob/OKTA-295020-content-types/src/Okta.Sdk.UnitTests/PayloadHandlerShould.cs#L18)
[x] assert that PayloadHandler.SetHttpRequestMessageContent sets Content-Transfer-Encoding header (https://github.com/okta/okta-sdk-dotnet/blob/OKTA-295020-content-types/src/Okta.Sdk.UnitTests/PayloadHandlerShould.cs#L34)

README.md
```diff
@@ -92,6 +92,19 @@ var client = new OktaClient(new OktaClientConfiguration
...
+### Create a scoped client
+Create a client scoped to a specific context to specify a custom content type:
+``` csharp
+var client = new OktaClient(new OktaClientConfiguration
+{
+    OktaDomain = "https://{{yourOktaDomain}}",
+    Token = "{{yourApiToken}}"
+});
+
+var scopedClient = client.CreatedScoped(new RequestContext { ContentType = "my-custom-content-type" });
```
laura-rodriguez, line 104: Based on our conversation last time, should we rename `CreatedScoped` method to something like `CreateScopedClient`?

bryanapellanes-okta: I think yes, we should.  I'll update shortly.

src/Okta.Sdk.UnitTests/OktaClientShould.cs
```diff
@@ -344,5 +342,12 @@ public async Task RetrieveNestedListOfObjects()
             statements.First().Type.Should().Be("EXPRESSION");
             statements.First().Name.Should().Be("Attribute");
         }
+
+        [Fact]
+        public void RegisterPkixCertPayloadHandlerOnConstruction()
+        {
+            _ = new OktaClient();
```
laura-rodriguez, line 349: Is this necessary?

src/Okta.Sdk.UnitTests/PayloadHandlerShould.cs
```diff
@@ -0,0 +1,78 @@
...
+            var testPayloadHandler = new TestPayloadHandler();
+            // register the payload handler so that it is found when the ContentType of the request is set. The OktaClient will register payload handlers as appropriate.
+            PayloadHandler.TryRegister(testPayloadHandler);
+            // set the ContentType of the request to match the ContentType of the TestPayloadHandler so that the request uses the TestPayloadHandler
+            var testHttpRequest = new HttpRequest { ContentType = "foo" };
+
+            var httpRequestMessage = new HttpRequestMessage();
+            httpRequestMessage.Content.Should().BeNull();
+            testPayloadHandler.SetHttpRequestMessageContent(testHttpRequest, httpRequestMessage);
+            httpRequestMessage.Content.Should().NotBeNull();
```
laura-rodriguez, line 30: Should we also assert the expected content?

src/Okta.Sdk/ApplicationsClient.cs
```diff
@@ -57,6 +58,88 @@ public Task<IApplication> CreateApplicationAsync(CreateBasicAuthApplicationOptio
...
         }
 
+        /// <inheritdoc />
+        public async Task<IJsonWebKey> PublishCerCertAsync(string base64EncodedCertificateData, string appId, string csrId, CancellationToken cancellationToken = default(CancellationToken))
+            => await PostAsync<JsonWebKey>(
+                new HttpRequest
+                {
+                    Uri = "/api/v1/apps/{appId}/credentials/csrs/{csrId}/lifecycle/publish",
+                    Verb = HttpVerb.Post,
+                    Payload = Convert.FromBase64String(base64EncodedCertificateData),
```
laura-rodriguez, line 68: Sanity check - Is this conversion necessary? This request sample seems to be passing a string: https://developer.okta.com/docs/reference/api/apps/#request-example-45

src/Okta.Sdk/ApplicationsClient.cs
```diff
@@ -57,6 +58,88 @@ public Task<IApplication> CreateApplicationAsync(CreateBasicAuthApplicationOptio
...
+
+        /// <inheritdoc />
+        public async Task<IJsonWebKey> PublishBinaryPemCertAsync(byte[] certificate, string appId, string csrId, CancellationToken cancellationToken = default(CancellationToken))
+            => await PostAsync<JsonWebKey>(
+                new HttpRequest
+                {
+                    Uri = "/api/v1/apps/{appId}/credentials/csrs/{csrId}/lifecycle/publish",
+                    Verb = HttpVerb.Post,
+                    Payload = certificate,
+                    ContentType = "application/pkix-cert",
```
laura-rodriguez, line 135: As per docs, we should use `application/x-pem-file` for PEM, see this example: https://developer.okta.com/docs/reference/api/apps/#request-example-45

src/Okta.Sdk/ApplicationsClient.cs
```diff
@@ -57,6 +58,88 @@ public Task<IApplication> CreateApplicationAsync(CreateBasicAuthApplicationOptio
...
+
+        /// <inheritdoc />
+        public async Task<IJsonWebKey> PublishBinaryDerCertAsync(byte[] certificate, string appId, string csrId, CancellationToken cancellationToken = default(CancellationToken))
+            => await PostAsync<JsonWebKey>(
+                new HttpRequest
+                {
+                    Uri = "/api/v1/apps/{appId}/credentials/csrs/{csrId}/lifecycle/publish",
+                    Verb = HttpVerb.Post,
+                    Payload = certificate,
+                    ContentType = "application/pkix-cert",
```
laura-rodriguez, line 119: Should we use `application/x-x509-ca-cert` for binary? See request examples here: https://developer.okta.com/docs/reference/api/apps/#request-example-45

src/Okta.Sdk/Generated/ApplicationsClient.Generated.cs
```diff
@@ -156,6 +156,7 @@ public async Task<ICsr> GetCsrForApplicationAsync(string appId, string csrId, Ca
                 },
                 }, cancellationToken).ConfigureAwait(false);
         
+        /*
         /// <inheritdoc />
```
laura-rodriguez, line 160: Do we need this commented code?

src/Okta.Sdk/Internal/HttpRequestMessageProvider.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
```
laura-rodriguez, line 1: Missing header?

src/Okta.Sdk/Internal/HttpRequestMessageProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            { HttpVerb.Get, HttpMethod.Get },
+            { HttpVerb.Post, HttpMethod.Post },
+            { HttpVerb.Put, HttpMethod.Put },
+            { HttpVerb.Delete, HttpMethod.Delete },
+        };
+
+        public static IHttpRequestMessageProvider Default => new HttpRequestMessageProvider();
+
+        public HttpRequestMessage CreateHttpRequestMessage(HttpRequest httpRequest, string relativePath)
```
laura-rodriguez, line 20: Nit: Let's add comments for public methods.

src/Okta.Sdk.UnitTests/OktaClientShould.cs
bryanapellanes-okta: The underscore is a recommendation from Visual Studio (https://docs.microsoft.com/en-us/dotnet/csharp/discards).  The test asserts that construction of the client registers the payload handler.

src/Okta.Sdk.UnitTests/PayloadHandlerShould.cs
bryanapellanes-okta: I'll add additional tests to validate content.

src/Okta.Sdk/ApplicationsClient.cs
bryanapellanes-okta: I may have misunderstood the requirements of the api.  I'll revisit.

src/Okta.Sdk/ApplicationsClient.cs
bryanapellanes-okta: You're right, yes, we should.  I'll revisit.

src/Okta.Sdk/Generated/ApplicationsClient.Generated.cs
bryanapellanes-okta: Nope, forgot to delete it

src/Okta.Sdk/Internal/HttpRequestMessageProvider.cs
bryanapellanes-okta: Good catch 👍

src/Okta.Sdk/ApplicationsClient.cs
bryanapellanes-okta: Good catch, thanks 👍

src/Okta.Sdk.UnitTests/PayloadHandlerShould.cs
bryanapellanes-okta: Added additional tests in this file.

src/Okta.Sdk/ApplicationsClient.cs
bryanapellanes-okta: Corrected implementation.

src/Okta.Sdk/ApplicationsClient.cs
```diff
@@ -65,9 +65,8 @@ public async Task<IJsonWebKey> PublishCerCertAsync(string base64EncodedCertifica
                 {
                     Uri = "/api/v1/apps/{appId}/credentials/csrs/{csrId}/lifecycle/publish",
                     Verb = HttpVerb.Post,
-                    Payload = Convert.FromBase64String(base64EncodedCertificateData),
+                    Payload = base64EncodedCertificateData,
                     ContentType = "application/pkix-cert",
-                    ContentTransferEncoding = "base64",
```
bryanapellanes-okta: The encoding gets set by the payload handler.

src/Okta.Sdk.UnitTests/OktaClientShould.cs
laura-rodriguez: Oh I see! you call the constructor to register the payloads.

src/Okta.Sdk.UnitTests/PayloadHandlerShould.cs
laura-rodriguez: Is it possible to assert that the content-type header of the content is the one we are expecting? Something like this:

```
httpRequestMessage.Content.ContentType.Should().Be("application/pkix-cert")
```
Unless we have other tests that already cover this assert.

bryanapellanes-okta:
Merging as is; known issue will be addressed in follow up.

PR #398, opened 2020-06-25
bytearraypayloadhandler properly sets content-type header

bryanapellanes-okta:
## Issue \#
When using Publish* Certificate methods that accept byte arrays (byte[]), content related headers are not correctly set.

## Code

- [x] Unit test(s)
- [x] Implementation

## Current behavior
When using Publish* Certificate methods that accept byte arrays (byte[]) content related headers are not correctly set.

## Desired behavior
Content related headers are correctly set.

## Additional Context
Confusion about the requirements and expectations of the publish certificate methods, that accept binary data, lead to an incomplete implementation.

Manually tested each of the methods that accept binary data and received no content header related errors.

### DIFF
--- a/src/Okta.Sdk.UnitTests/PayloadHandlerShould.cs
+++ b/src/Okta.Sdk.UnitTests/PayloadHandlerShould.cs
@@ -3,6 +3,7 @@
 // Licensed under the Apache 2.0 license. See the LICENSE file in the project root for full license information.
 // </copyright>
 
+using System.Linq;
 using System.Net.Http;
 using System.Text;
 using System.Threading.Tasks;
@@ -97,6 +98,8 @@ namespace Okta.Sdk.UnitTests
             pkixCertPayloadHandler.SetHttpRequestMessageContent(testHttpRequest, httpRequestMessage);
             httpRequestMessage.Content.Should().NotBeNull();
             httpRequestMessage.Content.GetType().Should().Be(typeof(StringContent));
+            httpRequestMessage.Content.Headers.Contains("Content-Type");
+            httpRequestMessage.Content.Headers.ContentType.ToString().Should().Be("application/pkix-cert; charset=utf-8");
         }
 
         [Fact]
@@ -115,6 +118,8 @@ namespace Okta.Sdk.UnitTests
             pkixCertPayloadHandler.SetHttpRequestMessageContent(testHttpRequest, httpRequestMessage);
             httpRequestMessage.Content.Should().NotBeNull();
             httpRequestMessage.Content.GetType().Should().Be(typeof(ByteArrayContent));
+            httpRequestMessage.Content.Headers.Contains("Content-Type");
+            httpRequestMessage.Content.Headers.ContentType.ToString().Should().Be("application/x-pem-file");
         }
 
         [Fact]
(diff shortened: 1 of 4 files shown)


src/Okta.Sdk/Internal/PkixCertPayloadHandler.cs
```diff
@@ -41,7 +41,17 @@ protected override HttpContent GetRequestHttpContent(HttpRequest httpRequest)
...
+            if (httpRequest.Payload is string stringPayload)
+            {
+                content = stringPayload;
+            }
+            else if (httpRequest.Payload is byte[] bytePayload)
+            {
+                content = Convert.ToBase64String(bytePayload);
+            }
+
+            return new StringContent(content, Encoding.UTF8, ContentType);
```
bryanapellanes-okta, line 54: Because this payload handler can accept either base64 encoded string or byte array (byte[]), I had to account for that here.

src/Okta.Sdk.UnitTests/PayloadHandlerShould.cs
```diff
@@ -115,6 +118,8 @@ public void SetPemCertContent()
             pkixCertPayloadHandler.SetHttpRequestMessageContent(testHttpRequest, httpRequestMessage);
             httpRequestMessage.Content.Should().NotBeNull();
             httpRequestMessage.Content.GetType().Should().Be(typeof(ByteArrayContent));
+            httpRequestMessage.Content.Headers.Contains("Content-Type");
+            httpRequestMessage.Content.Headers.ContentType.ToString().Should().Be("application/x-pem-file");
```
laura-rodriguez, line 122: +1 to this addition! This is was I was talking about yesterday. Thanks for adding this.

src/Okta.Sdk/Internal/PkixCertPayloadHandler.cs
```diff
@@ -41,7 +41,17 @@ protected override HttpContent GetRequestHttpContent(HttpRequest httpRequest)
...
 
-            return new StringContent((string)httpRequest.Payload, Encoding.UTF8, ContentType);
+            var content = string.Empty;
+            if (httpRequest.Payload is string stringPayload)
+            {
+                content = stringPayload;
+            }
+            else if (httpRequest.Payload is byte[] bytePayload)
+            {
+                content = Convert.ToBase64String(bytePayload);
```
laura-rodriguez, line 51: Sanity-check question: Is this conversion necessary when you are passing a byte[]? Is this how the request looks in our docs?

src/Okta.Sdk.UnitTests/PayloadHandlerShould.cs
bryanapellanes-okta: I was confused about how to correctly do this assertion, I figured it out thankfully.

src/Okta.Sdk/Internal/PkixCertPayloadHandler.cs
bryanapellanes-okta: Yes, this payload handler handles content for the Content-Type "application/pkix-cert" and will always send the payload as a base64 encoded string.  So, if a byte array is provided, that byte array needs to be converted to base64 in order for the headers in the message to be valid.

PR #401, opened 2020-07-01
OKTA-303288: add coverage script; fix tests for latest openapi regen

bryanapellanes-okta:
## Issue \#
[OKTA-303288](https://oktainc.atlassian.net/browse/OKTA-303288)

## Code

- [x] Unit test(s)
- [ ] Implementation

## Additional Context
Added a script used to get code coverage report from integration and unit test runs.  Discovered a couple of test failures where the test code needed updates to align with latest openapi regenerated code.

After running:
```sh
./coverage.sh
```

You can open `./coverage-html/index.html` to view results.


src/Okta.Sdk.IntegrationTests/xunit.runner.json
```diff
@@ -1,5 +1,5 @@
 ﻿{
-  "parallelizeAssembly": true,
-  "parallelizeTestCollections": true,
+  "parallelizeAssembly": false,
+  "parallelizeTestCollections": false,
```
laura-rodriguez, line 3: Nice! Since it's possible to configure this via JSON for the entire project, should we remove the configuration we added recently in the code?

bryanapellanes-okta: I'm hesitant to do that because it seems the behavior is dependent on the runner used for a given run.

PR #404, opened 2020-07-06
update version from 2.0.0 to 3.0.0

bryanapellanes-okta:


laura-rodriguez:
Project version looks good. We should also update the [readme](https://github.com/okta/okta-sdk-dotnet/#release-status) including the new version, and the retiring warning for 2.0.0 which should be in 9 months (more info [here](https://developer.okta.com/code/library-versions/#retiring)).

PR #409, opened 2020-07-14
Added support for Password Import Inline Hook user creation.

### LINKED ISSUE #408: [Request] Create User With Password Import Inline Hook
## User story

As a developer, I want to be able to create users with [password import inline hook](https://developer.okta.com/docs/reference/password-hook/), so that I can easily develop an user importation solution.

## Proposed solution
Either a new options class such as `CreateActiveUserWithInlineHook` and a corresponding `IUsersClient.CreateUserAsync` to go with it, or by altering the [IPasswordCredential](https://github.com/okta/okta-sdk-dotnet/blob/master/src/Okta.Sdk/Generated/IPasswordCredential.Generated.cs) interface to support a [Password Import Inline Hook](https://developer.okta.com/docs/reference/api/users/#create-user-with-password-import-inline-hook)

## Alternatives considered
Using a `HttpRequest` directly.

HumanAftrAll:
## Issue \#
Fixes #408 

## Code

- [ ] Unit test(s)
- [x] Implementation

## Current behavior
Lacks support for Password Import Inline Hook user creation.

## Desired behavior
Add support for Password Import Inline Hook user creation.


laura-rodriguez:
This code was included in #410 .

PR #416, opened 2020-08-03
OKTA-316815 expose trusted origins client

### LINKED ISSUE #413: Unable to access the TrustedOriginsClient from IOktaClient
I was looking to use the `TrustedOriginsClient` and see it was added in 3.0, but unfortunately it does not seem currently accessible from `IOktaClient`.

As a workaround, I am able to do everything I need using the appropriate `GetAsync`, `PostAsync<T>`, etc. methods on `IOktaClient`, which works great! It's been especially easy since the `ITrustedOrigin` and `TrustedOrigin` objects already exist for me to use.

## Current behavior
Unable to access the `TrustedOriginsClient` from `IOktaClient`.

## Expected behavior
Make calls to the `TrustedOriginsClient` by using `IOktaClient`.

## Environment

- OS:
- Browser: 
- .Net Framework: .Net Core 3.1
- Other:

bryanapellanes-okta:
## Issue \#
This  change adds the TrustedOrigins client to the default OktaClient class definition.
[OKTA-316815](https://oktainc.atlassian.net/browse/OKTA-316815)

Closes #413

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/TrustedOriginScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/TrustedOriginScenarios.cs
@@ -21,16 +21,8 @@ namespace Okta.Sdk.IntegrationTests
         public async Task ListOrigins()
         {
             var client = TestClient.Create();
-            var trustedOrigins = await client.TrustedOrigins.ListOrigins().ToListAsync();
-            trustedOrigins.Should().NotBeNull();
-            trustedOrigins.Count.Should().BeGreaterThan(0);
-        }
-
-        [Fact]
-        public async Task CreateOrigin()
-        {
-            var client = TestClient.Create();
-            var testTrustedOriginName = $"{dotnetSdkPrefix}_{nameof(CreateOrigin)}_Test";
+            var guid = Guid.NewGuid();
+            var testTrustedOriginName = $"{dotnetSdkPrefix}_{nameof(ListOrigins)}_{guid}_Test";
             var createdTrustedOrigin = await client.TrustedOrigins.CreateOriginAsync(
                 new TrustedOrigin
                 {
@@ -40,11 +32,48 @@ namespace Okta.Sdk.IntegrationTests
                     {
                         new Scope()
                         {
-                            Type = "CORS",
+                            Type = ScopeType.Cors,
                         },
                         new Scope()
                         {
-                            Type = "REDIRECT",
+                            Type = ScopeType.Redirect,
+                        },
+                    },
+                });
+
+            try
+            {
+                var trustedOrigins = await client.TrustedOrigins.ListOrigins().ToListAsync();
+                trustedOrigins.Should().NotBeNull();
+                trustedOrigins.Count.Should().BeGreaterThan(0);
+                trustedOrigins.FirstOrDefault(to => to.Name.Equals(testTrustedOriginName)).Should().NotBeNull();
+            }
+            finally
+            {
+                await client.TrustedOrigins.DeleteOriginAsync(createdTrustedOrigin.Id);
+            }
+        }
+
+        [Fact]
+        public async Task CreateOrigin()
+        {
+            var client = TestClient.Create();
+            var guid = Guid.NewGuid();
+            var testTrustedOriginName = $"{dotnetSdkPrefix}_{nameof(CreateOrigin)}_{guid}_Test";
+            var createdTrustedOrigin = await client.TrustedOrigins.CreateOriginAsync(
+                new TrustedOrigin
+                {
+                    Name = testTrustedOriginName,
+                    Origin = "http://example.com",
+                    Scopes = new List<IScope>()
+                    {
+                        new Scope()
+                        {
+                            Type = ScopeType.Cors,
+                        },
+                        new Scope()
+                        {
+                            Type = ScopeType.Redirect,
                         },
                     },
                 });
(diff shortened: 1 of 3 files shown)


src/Okta.Sdk.IntegrationTests/TrustedOriginScenarios.cs
```diff
@@ -0,0 +1,241 @@
...
+{
+    public class TrustedOriginScenarios
+    {
+        private static string dotnetSdkPrefix = "dotnetSdk";
+
+        [Fact]
+        public async Task ListOrigins()
+        {
+            var client = TestClient.Create();
+            var trustedOrigins = await client.TrustedOrigins.ListOrigins().ToListAsync();
```
laura-rodriguez, line 24: To test this scenario, let's create a Trusted Origin and verify if it is listed by this method. We shouldn't rely on existing origins.

src/Okta.Sdk.IntegrationTests/TrustedOriginScenarios.cs
```diff
@@ -0,0 +1,241 @@
...
+            trustedOrigins.Should().NotBeNull();
+            trustedOrigins.Count.Should().BeGreaterThan(0);
+        }
+
+        [Fact]
+        public async Task CreateOrigin()
+        {
+            var client = TestClient.Create();
+            var testTrustedOriginName = $"{dotnetSdkPrefix}_{nameof(CreateOrigin)}_Test";
```
laura-rodriguez, line 33: Do we need a random keyword to avoid name collisions?

src/Okta.Sdk.IntegrationTests/TrustedOriginScenarios.cs
```diff
@@ -0,0 +1,241 @@
...
+            var createdTrustedOrigin = await client.TrustedOrigins.CreateOriginAsync(
+                new TrustedOrigin
+                {
+                    Name = testTrustedOriginName,
+                    Origin = "http://example.com",
+                    Scopes = new List<IScope>()
+                    {
+                        new Scope()
+                        {
+                            Type = "CORS",
```
laura-rodriguez, line 43: Does this work if we pass `ScopeType.Cors` instead?

src/Okta.Sdk.IntegrationTests/TrustedOriginScenarios.cs
```diff
@@ -0,0 +1,241 @@
...
+            {
+                await client.TrustedOrigins.DeleteOriginAsync(createdTrustedOrigin.Id);
+            }
+        }
+
+        [Fact]
+        public async Task DeleteOrigin()
+        {
+            var client = TestClient.Create();
+            var testTrustedOriginName = $"{dotnetSdkPrefix}_{nameof(DeleteOrigin)}_Test";
```
laura-rodriguez, line 70: Do we need a random keyword to avoid name collisions?

src/Okta.Sdk.IntegrationTests/TrustedOriginScenarios.cs
```diff
@@ -0,0 +1,241 @@
...
+            await client.TrustedOrigins.DeleteOriginAsync(createdTrustedOrigin.Id);
+
+            var ex = await Assert.ThrowsAsync<OktaApiException>(() => client.TrustedOrigins.GetOriginAsync(createdTrustedOrigin.Id));
+            ex.StatusCode.Should().Be(404);
+        }
+
+        [Fact]
+        public async Task GetOrigin()
+        {
+            var client = TestClient.Create();
```
laura-rodriguez, line 100: To test this scenario, let's create a Trusted Origin and verify if it is retrieved by `GetOriginAsync`. We shouldn't rely on existing origins.

src/Okta.Sdk.IntegrationTests/TrustedOriginScenarios.cs
```diff
@@ -0,0 +1,241 @@
...
+            retrievedTrustedOrigin.Id.Should().Be(trustedOrigin.Id);
+            retrievedTrustedOrigin.Name.Should().Be(trustedOrigin.Name);
+            retrievedTrustedOrigin.Origin.Should().Be(trustedOrigin.Origin);
+        }
+
+        [Fact]
+        public async Task UpdateOrigin()
+        {
+            var client = TestClient.Create();
+            var testTrustedOriginName = $"{dotnetSdkPrefix}_{nameof(UpdateOrigin)}_Test";
```
laura-rodriguez, line 116: Do we need a random keyword to avoid name collisions?

src/Okta.Sdk.IntegrationTests/TrustedOriginScenarios.cs
```diff
@@ -0,0 +1,241 @@
...
+            {
+                await client.TrustedOrigins.DeleteOriginAsync(createdTrustedOrigin.Id);
+            }
+        }
+
+        [Fact]
+        public async Task ActivateOrigin()
+        {
+            var client = TestClient.Create();
+            var testTrustedOriginName = $"{dotnetSdkPrefix}_{nameof(ActivateOrigin)}_Test";
```
laura-rodriguez, line 173: Do we need a random keyword to avoid name collisions?

src/Okta.Sdk.IntegrationTests/TrustedOriginScenarios.cs
bryanapellanes-okta: Good idea; my thought was that would only be necessary if the finally block below was somehow not reached.  However, multiple instances of this test may run simultaneously so I will put in the random keyword as you suggest.

bryanapellanes-okta:
@laura-rodriguez implemented your suggestions:
- randomize names to avoid collisions (using guid)
- use ScopeType enum values
- don't rely on existing values, instead create new and verify

PR #417, opened 2020-08-10
OKTA-317892: run test with retry to address flakiness

bryanapellanes-okta:
## Issue \#
~~Added minimal retry strategy to re-attempt flaky test 2 more times (in addition to the initial attempt, for a total of 3 tries).~~
Changed xunit config to use single thread.
[OKTA-317892](https://oktainc.atlassian.net/browse/OKTA-317892)

## Code

- [ ] Unit test(s)
- [ ] Implementation

## Current behavior
Test fails intermittently,

## Desired behavior
Test should always pass

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/UserScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/UserScenarios.cs
@@ -5,7 +5,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Okta.Sdk.IntegrationTests/UserScenarios.cs
```diff
@@ -763,37 +764,40 @@ public async Task RemoveGroupTargetFromRole()
...
-                forgotPasswordResponse.ResetPasswordUrl.Should().NotBeNullOrEmpty();
-            }
-            finally
-            {
-                await createdUser.DeactivateAsync();
-                await createdUser.DeactivateOrDeleteAsync();
-            }
+                try
+                {
+                    var forgotPasswordResponse = await createdUser.ForgotPasswordGenerateOneTimeTokenAsync(false);
```
laura-rodriguez, line 791: +1 for the retry approach, but I think we should only retry 
`var forgotPasswordResponse = await createdUser.ForgotPasswordGenerateOneTimeTokenAsync(false);` plus asserts since that piece of code seems to be the problematic one. User creation shouldn't be retried.

We have done something similar in the past, so FYI you can use `Polly` that is already installed and does the retry work for you: https://github.com/okta/okta-sdk-dotnet/blob/cb0b36c4732ef75967a3c2d1bca1ea9c766a5ead/src/Okta.Sdk.IntegrationTests/UserScenarios.cs#L60

PR #422, opened 2020-09-07
OKTA-324648 .NET Management SDK > Exception when using the SDK on AWS…

andriizhegurov-okta:
… lambda

Added `TryGetHomePath` function that won't throw an exception if HOME environment variable is missing (this is the case for AWS Serverless environment)
`Resolve` function is changed to use `TryGetHomePath` instead of `GetHomePath`.
If `TryGetHomePath` fails, Resolve returns null what is a valid parameter value for FlexibleConfiguration's `ConfigurationBuilder`.

## Issue \#

https://oktainc.atlassian.net/browse/OKTA-324648

## Code

- [ ] Unit test(s)
- [x] Implementation

## Current behavior

In a Linux environment when HOME environment variable isn't defined (no access to home directory), OktaClient can't be created because of exception in `Home.GetPath` function. Issue was detected in AWS lambda / serverless environment.

## Desired behavior

No exceptions thrown in the above-mentioned case because there are other possible configuration sources supported by OktaClient besides user home directory.

### DIFF
--- a/src/Okta.Sdk/Internal/HomePath.cs
+++ b/src/Okta.Sdk/Internal/HomePath.cs
@@ -22,7 +22,7 @@ namespace Okta.Sdk.Internal
         /// Provides support for Unix-like paths on Windows. If the first path segment starts with <c>~</c>, this segment is prepended with the home directory path.
         /// </remarks>
         /// <param name="pathSegments">The path segments.</param>
-        /// <returns>A combined path which includes the resolved home directory path (if necessary).</returns>
+        /// <returns>A combined path which includes the resolved home directory path (if necessary). If home directory path cannot be resolved, returns null.</returns>
         public static string Resolve(params string[] pathSegments)
         {
             if (pathSegments.Length == 0)
@@ -35,7 +35,10 @@ namespace Okta.Sdk.Internal
                 return Path.Combine(pathSegments);
             }
 
-            var homePath = GetHomePath();
+            if (!TryGetHomePath(out string homePath))
+            {
+                return null;
+            }
 
             var newSegments =
                 new string[] { pathSegments[0].Replace("~", homePath) }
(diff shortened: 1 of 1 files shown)


bryanapellanes-okta (approved):
This looks fine to me.

src/Okta.Sdk/Internal/HomePath.cs
```diff
@@ -47,28 +50,43 @@ public static string Resolve(params string[] pathSegments)
 
         /// <summary>
         /// Resolves the current user's home directory path.
+        /// Throws an exception if the path cannot be resolved.
         /// </summary>
         /// <returns>The home path.</returns>
         public static string GetHomePath()
         {
```
laura-rodriguez, line 57: Is this method still necessary?

laura-rodriguez:
Thanks for the details in the PR description @andriizhegurov-okta ! Much appreciated.
So, what's the recommended configuration option when using AWS lambda? inline/yaml file? I think we can add a section in the readme describing this.

laura-rodriguez (commented):
Is it possible to add some unit tests? Not sure how complicated is to mock env vars.

src/Okta.Sdk/Internal/HomePath.cs
andriizhegurov-okta: This one is declared as public so this would be breaking changes if we remove it. 
I asked Bryan about this and he said if we decide to delete a public function, we need to follow a deprecation procedure (obsolete things for deletion in a later release).
Should we do this, what do you think?

andriizhegurov-okta:
> So, what's the recommended configuration option when using AWS lambda? inline/yaml file? I think we can add a section in the readme describing this.

This is a really good question. I'm not an AWS expert, so far I discovered that configuration can be fetched from `appsettings.json` or `okta.yaml` from the current directory. 
Also I may be wrong but I suspect that an access to home directory can be granted to user using AWS policies and environment variables can be set though they are not set initially. This is better to be investigated, we may do it in scope of another ticket and then add instructions about how to enable this access and set the environment var.

laura-rodriguez:
> > So, what's the recommended configuration option when using AWS lambda? inline/yaml file? I think we can add a section in the readme describing this.
> 
> This is a really good question. I'm not an AWS expert, so far I discovered that configuration can be fetched from `appsettings.json` or `okta.yaml` from the current directory.
> Also I may be wrong but I suspect that an access to home directory can be granted to user using AWS policies and environment variables can be set though they are not set initially. This is better to be investigated, we may do it in scope of another ticket and then add instructions about how to enable this access and set the environment var.
> 
> Update: Or we may skip those investigations and just mention in the first line below something like "some environments like AWS Lambda may not provide an access to user's home directory by default"
> image

So, If they provide either yaml or inline configuration it won't fail, right? As soon as the SDK discovers that there is not an env var for this will try to get it from other sources. Is that correct? If that's the case then I think we are covered.

src/Okta.Sdk/Internal/HomePath.cs
laura-rodriguez: I hope no one is using this since this is in the `Internal` namespace. However, let's keep it just in case and we can remove it in the next major version.

andriizhegurov-okta:
> > 
> 
> So, If they provide either yaml or inline configuration it won't fail, right? As soon as the SDK discovers that there is not an env var for this will try to get it from other sources. Is that correct? If that's the case then I think we are covered.

@laura-rodriguez 
Right, now after this fix it will work the way you described. But before the fix `HomePath.Resolve` threw an exception if it could not find HOME environment variable (and this is not right because `homeOktaYamlLocation` is used as an optional configuration source).
Now in this case `homeOktaYamlLocation` will be null and `ConfigurationBuilder` will skip it.
Here is the code:
image

laura-rodriguez (approved):

PR #430, opened 2020-09-14
OKTA-325664

andriizhegurov-okta:
.NET Management SDK > Verify that `AddFactorAsync` overloads are up to date

## Issue \#

https://oktainc.atlassian.net/browse/OKTA-325664

## Code

- [x] Implementation

- [x] Unit test(s)

## Current behavior

Not possible to specify `tokenLifetimeSeconds` on enrolling an email factor via SDK though it's mentioned in the docs.
https://developer.okta.com/docs/reference/api/factors/#enroll-okta-email-factor

## Desired behavior

Need to add this parameter and check for other discrepancies.

## Additional Context

Following issues were found while reviewing the documentation (https://developer.okta.com/docs/reference/api/factors/#factor-lifecycle-operations)
* Wrong link for `Upload YubiKey Seed` https://developer.okta.com/docs/reference/api/factors/#upload-yubikey-seed instead of https://developer.okta.com/docs/reference/api/factors/#upload-yubikey-otp-seed
* Providers and factors from there https://developer.okta.com/docs/reference/api/factors/#provider-type don't fully match this list https://developer.okta.com/docs/reference/api/factors/#factor-lifecycle-operations - this needs to be addressed. After that we'll need to implement helpers for missing factors in SDK.

### DIFF
--- a/src/Okta.Sdk.UnitTests/OktaClientShould.cs
+++ b/src/Okta.Sdk.UnitTests/OktaClientShould.cs
@@ -4,7 +4,6 @@
 // </copyright>
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
 using System.Threading;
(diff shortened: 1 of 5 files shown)


src/Okta.Sdk.UnitTests/OktaClientShould.cs
```diff
@@ -354,5 +355,35 @@ public void RegisterPayloadHandlers()
             PayloadHandler.ForContentType("application/x-pem-file").GetType().Should().Be(typeof(PemFilePayloadHandler));
             PayloadHandler.ForContentType("application/x-x509-ca-cert").GetType().Should().Be(typeof(X509CaCertPayloadHandler));
         }
+
+        [Fact]
+        public async Task EnrollEmailFactor()
```
laura-rodriguez, line 360: Can you please create a separate class called `UserFactorsClientShould` and move factors tests there? That way is easy to find a test for a particular client. We did the same for [`ApplicationsClient`](https://github.com/okta/okta-sdk-dotnet/blob/master/src/Okta.Sdk.UnitTests/ApplicationsClientShould.cs) for example.

andriizhegurov-okta: Done

src/Okta.Sdk/AddEmailFactorOptions.cs
```diff
@@ -14,5 +14,10 @@ public sealed class AddEmailFactorOptions
         /// Gets or sets the email
         /// </summary>
         public string Email { get; set; }
+
+        /// <summary>
+        /// Gets or sets the lifetime of the OTP in a range 1 to 86400, 300 seconds is the default.
```
laura-rodriguez, line 19: Instead of describing the range in the comments, I suggest we add a [`Remarks`](https://docs.microsoft.com/en-us/dotnet/csharp/codedoc#remarks) (or `<see>`/`<seealso>`) section with a link to the docs (for example, `<Remarks>For more details see [link]</Remarks>`. Since our team doesn't own this API, if the API team decides to change the range we will need to release a new version just to update the comments. This is a general rule for methods/attributes documentation. 
Defaults are OK since they generally explicitly set in our code.

src/Okta.Sdk.UnitTests/UserFactorsClientShould.cs
```diff
@@ -0,0 +1,44 @@
...
+            var factorsClient = client.UserFactors;
+            var emailFactorOptions = new AddEmailFactorOptions
+            {
+                Email = "johndoe@mail.com",
+                TokenLifetimeSeconds = 999,
+            };
+            await factorsClient.AddFactorAsync("UserId", emailFactorOptions);
+            await mockDataStore
+                .Received(1)
+                .PostAsync<UserFactor>(Arg.Is<HttpRequest>(request => RequestMatchOptions(request, emailFactorOptions)), Arg.Any<RequestContext>(), Arg.Any<CancellationToken>());
```
laura-rodriguez, line 28: For testing in general, we usually favor readability over reusability. The idea is that tests should be easy to read and follow by reviewers (especially when we add a new API because the reviewer has to take a look at a considerable number of tests, and we should facilitate that process). You can see this pattern in _almost_ all our tests. 

I suggest we inline this logic here instead of using a separate private method. Also, in addition to verifying that the query params are correct, we should check not only that the payload has all the fields we set but also the structure (correct nesting, etc.).

laura-rodriguez (commented):
It looks good overall. I left some minor comments (mostly nit-picks).

src/Okta.Sdk.UnitTests/UserFactorsClientShould.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
```
laura-rodriguez, line 1: Check StyleCop warnings, since IIRC it will ask you to add a header here.

laura-rodriguez (approved):
LGTM

laura-rodriguez:
Superseded by #431.

PR #434, opened 2020-09-23
OKTA-325665 .NET Management SDK > Add helper methods for Policy

### LINKED ISSUE #287: Fix all TODOs from policy module.
There are a couple of helpers that would be nice to be added. They are marked as TODOs in the code.

andriizhegurov-okta:
## Issue \#

https://oktainc.atlassian.net/browse/OKTA-325665

Fix #287 

## Code

- [x] Unit test(s)
- [x] Implementation

## Desired behavior

Resolved several TODOs: added helper classes/functions and adjusted ITs accordingly

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/PoliciesScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/PoliciesScenarios.cs
@@ -407,9 +407,9 @@ namespace Okta.Sdk.IntegrationTests
                 NetworkConditionConnection = "ANYWHERE",
                 NetworkConditionInclude = new List<string>(),
                 NetworkConditionExclude = new List<string>(),
-                PasswordChangeAccess = "ALLOW",
-                SelfServicePasswordResetAccess = "ALLOW",
-                SelfServiceUnlockAccess = "DENY",
+                ActionPasswordChangeAccess = "ALLOW",
+                ActionSelfServicePasswordResetAccess = "ALLOW",
+                ActionSelfServiceUnlockAccess = "DENY",
             };
 
             var createdPolicyRule = await client.Policies.AddPolicyRuleAsync(policyRuleOptions);
@@ -451,14 +451,14 @@ namespace Okta.Sdk.IntegrationTests
             var policyRuleOptions = new AddSignOnPolicyRuleOptions(createdPolicy.Id)
             {
                 Name = $"dotnet-sdk: CreateOktaSignOnOnPremPolicyRule {guid}".Substring(0, 50),
-                Access = "ALLOW",
-                RequireFactor = false,
-                RememberDeviceByDefault = false,
-                FactorLifetime = 10,
-                UsePersistentCookie = false,
-                MaxSessionIdleMinutes = 720,
-                MaxSessionLifetimeMinutes = 0,
-                AuthType = "ANY",
+                SignonActionAccess = "ALLOW",
+                SignonActionRequireFactor = false,
+                SignonActionRememberDeviceByDefault = false,
+                SignonActionFactorLifetime = 10,
+                SignonSessionUsePersistentCookie = false,
+                SignonSessionMaxSessionIdleMinutes = 720,
+                SignonSessionMaxSessionLifetimeMinutes = 0,
+                AuthContextConditionAuthType = "ANY",
             };
 
             var createdPolicyRule = await client.Policies.AddPolicyRuleAsync(policyRuleOptions);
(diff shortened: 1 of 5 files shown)


src/Okta.Sdk.IntegrationTests/PoliciesScenarios.cs
```diff
@@ -396,50 +396,27 @@ public async Task CreatePasswordPolicyRule()
...
-                    SelfServicePasswordReset = new PasswordPolicyRuleAction()
-                    {
-                        Access = "ALLOW",
-                    },
-                    SelfServiceUnlock = new PasswordPolicyRuleAction()
-                    {
-                        Access = "DENY",
-                    },
-                },
+                UsersConditionExclude = new List<string>(),
```
laura-rodriguez, line 404: I wonder if we should add a prefix to the properties to indicate context, for example, instead of `UsersConditionInclude` rename it to `PeopleConditionUsersConditionInclude`? I believe it's clearer, you know exactly that the property is for PeopleConditions. Thoughts @andriizhegurov-okta @bryanapellanes-okta ?

andriizhegurov-okta: In this case it's more like `PeopleConditionUsersConditionsInclude` and so on, or in a more concise way `PeopleConditionsIncludeUsers` and other in the same class: `PeopleConditionsExcludeUsers`, `PeopleConditionsIncludeGroups`, `PeopleConditionsExcludeGroups`
I'm ok with this naming but since this is the only Users-to-Include list I'm not sure we these long prefixes are needed.

andriizhegurov-okta: It's more a matter of consistency. How do we usually name the properties in helpers?

laura-rodriguez: The idea is that mapping properties with our docs should be easy. Let's say I want to find this property in the [API reference](https://developer.okta.com/docs/reference/api/policy/#people-condition-object), as you can see it's under the `People object`, so I think a `People` prefix is helpful.

I like `PeopleConditionIncludeUsers` , I think it would be even more helpful if we add a `<remarks>` section in the comments with a link to the docs. For example:

```
/// <summary>
/// Gets or sets a list of users to be excluded
/// </summary>
<remarks>See <a href="https://developer.okta.com/docs/reference/api/policy/#people-condition-object"> the API documentation on policy objects. </remarks>
public IList<string> PeopleConditionIncludeUsers { get; set; }
```

andriizhegurov-okta: Done

src/Okta.Sdk/PoliciesClient.cs
```diff
@@ -12,33 +13,153 @@ namespace Okta.Sdk
...
             => await CreatePolicyAsync(policy, activate, cancellationToken).ConfigureAwait(false) as T;
 
         /// <inheritdoc/>
-        public async Task<T> UpdatePolicyAsync<T>(IPolicy policy, string policyId, CancellationToken cancellationToken = default(CancellationToken))
+        public async Task<T> UpdatePolicyAsync<T>(IPolicy policy, string policyId, CancellationToken cancellationToken = default)
             where T : class, IPolicy
             => await UpdatePolicyAsync(policy, policyId, cancellationToken).ConfigureAwait(false) as T;
 
         /// <inheritdoc/>
-        public async Task<T> AddPolicyRuleAsync<T>(IPolicyRule policyRule, string policyId, CancellationToken cancellationToken = default(CancellationToken))
```
laura-rodriguez, line 30: We shouldn't remove the existing method, otherwise, it's gonna be a breaking change.

laura-rodriguez (commented):
I love to see //TODOs gone 🎉 ! 
Please see my comment about breaking changes. Also, I suggest doing the same prefix + remarks strategy for all the nested properties.

src/Okta.Sdk/PoliciesClient.cs
andriizhegurov-okta: This method is not removed, I only added mine before it and simplified the signature:
`public async Task<T> AddPolicyRuleAsync<T>(IPolicyRule policyRule, string policyId, CancellationToken cancellationToken = default)`
instead of
`public async Task<T> AddPolicyRuleAsync<T>(IPolicyRule policyRule, string policyId, CancellationToken cancellationToken = default(CancellationToken ))`

despite it's shown as deleted, it's still there, couple lines further

laura-rodriguez (approved):
👍

bryanapellanes-okta (approved):
👍

PR #437, opened 2020-09-29
OKTA-331690

andriizhegurov-okta:
.NET > Management Sdk > Add overload to add HOTP factor

## Issue \#

https://oktainc.atlassian.net/browse/OKTA-331690

## Code

- [x] Unit test(s)
- [x] Implementation

## Additional Context

Added overloads and helper classes for HOTP Factor

### DIFF
--- a/src/Okta.Sdk/Generated/SamlApplicationSettingsSignOn.Generated.cs
+++ b/src/Okta.Sdk/Generated/SamlApplicationSettingsSignOn.Generated.cs
@@ -17,20 +17,6 @@ namespace Okta.Sdk
     public sealed partial class SamlApplicationSettingsSignOn : Resource, ISamlApplicationSettingsSignOn
     {
         /// <inheritdoc/>
-        public IList<IAcsEndpoint> AcsEndpoints 
-        {
-            get => GetArrayProperty<IAcsEndpoint>("acsEndpoints");
-            set => this["acsEndpoints"] = value;
-        }
-        
-        /// <inheritdoc/>
-        public bool? AllowMultipleAcsEndpoints 
-        {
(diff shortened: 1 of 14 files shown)


src/Okta.Sdk/Generated/SamlApplicationSettingsSignOn.Generated.cs
```diff
@@ -16,6 +16,20 @@ namespace Okta.Sdk
     /// <inheritdoc/>
     public sealed partial class SamlApplicationSettingsSignOn : Resource, ISamlApplicationSettingsSignOn
     {
+        /// <inheritdoc/>
+        public IList<IAcsEndpoint> AcsEndpoints
```
laura-rodriguez, line 20: Did these changes come with the latest version of the openapi spec?

andriizhegurov-okta: I shouldn't have included this in the PR, this is not related to the issue. I'll remove it, thanks.

laura-rodriguez: It's OK, this could happen when you run the generation script using the latest version of the spec. If changes are small, I'm OK including them on the same PR, otherwise, we should have a separate PR with the latest spec changes.

andriizhegurov-okta: Yes, there were several other changes to generated files not related to the issue, I excluded them from the initial PR but missed this one. 
I think we should not put code changes done in scope of different stories in one PR. So I removed this changes.

PR #445, opened 2020-10-20
Removed secure environment variable from .travis.yml

andriizhegurov-okta:
## Issue \#

https://oktainc.atlassian.net/browse/OKTA-324681

## Code

- [ ] Unit test(s)
- [x] Implementation

## Additional Context

Removed secure env variable from the configuration file so now the one from Travis config is effective

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,6 @@ deploy:
   on:
     condition: "$DEPLOY_DOCS = true && $TRAVIS_PULL_REQUEST = false"
     branch: master
-env:
-  global:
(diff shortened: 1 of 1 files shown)


bryanapellanes-okta (commented):
This seems fine, I'd be concerned that something else might be dependent on that value though.  I'd want to hear what @laura-rodriguez and @bdemers have to say.

bdemers (approved):

laura-rodriguez (approved):
Let's monitor that everything else is still working as expected.

PR #449, opened 2020-11-03
OKTA-341985 regenerate code with openapi spec 2.1.5

laura-rodriguez:
* Regenerate code using [openapi 2.1.5](http://npmjs.com/package/@okta/openapi/v/2.1.5). 
* Update ITs to cover latest changes.
* Add obsolete tag to `PasswordCredentialHash.WorkerFactor` and suggest using `PasswordCredentialHash.WorkFactor`.

## Issue \#
OKTA-341985

## Code

- [x] Unit test(s)
- [x] Implementation

### DIFF
--- a/src/Okta.Sdk/Generated/UserFactorResolver.Generated.cs
+++ b/src/Okta.Sdk/Generated/UserFactorResolver.Generated.cs
@@ -40,11 +40,6 @@ namespace Okta.Sdk.Internal
                 return typeof(EmailUserFactor);
             }
             
-            if (value.Equals("hotp"))
-            {
-                return typeof(CustomHotpUserFactor);
-            }
-            
             if (value.Equals("push"))
             {
                 return typeof(PushUserFactor);
@@ -68,6 +63,11 @@ namespace Okta.Sdk.Internal
             if (value.Equals("token:hardware"))
             {
                 return typeof(HardwareUserFactor);
+            }
+            
+            if (value.Equals("token:hotp"))
+            {
+                return typeof(CustomHotpUserFactor);
             }
             
             if (value.Equals("token:software:totp"))
--- a/src/Okta.Sdk/PasswordCredentialHash.cs
+++ b/src/Okta.Sdk/PasswordCredentialHash.cs
@@ -1,0 +1,25 @@
+﻿// <copyright file="PasswordCredentialHash.cs" company="Okta, Inc">
+// Copyright (c) 2020 - present Okta, Inc. All rights reserved.
+// Licensed under the Apache 2.0 license. See the LICENSE file in the project root for full license information.
+// </copyright>
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Okta.Sdk
+{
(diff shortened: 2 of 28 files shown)


src/Okta.Sdk/PasswordCredentialHash.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Okta.Sdk
+{
+    /// <inheritdoc/>
+    public partial class PasswordCredentialHash : Resource, IPasswordCredentialHash
+    {
+        [Obsolete("This method will deprecated in the next major version. Use WorkFactor instead.")]
+        /// <summary>
+        /// Gets and sets the <c>workFactor</c> property. 
+        /// </summary>
+        public int? WorkerFactor
```
andriizhegurov-okta, line 19: Shouldn't it be called "WorkFactor"?

laura-rodriguez: I added a partial class with this property to avoid a breaking change. As you can see, internally maps to `workFactor` and I also tagged it as obsolete.

laura-rodriguez: Forgot to mention, that the generation script creates a `WorkFactor` property in the generated class.

src/Okta.Sdk/PasswordCredentialHash.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Okta.Sdk
+{
+    /// <inheritdoc/>
+    public partial class PasswordCredentialHash : Resource, IPasswordCredentialHash
+    {
+        [Obsolete("This method will deprecated in the next major version. Use WorkFactor instead.")]
```
bryanapellanes-okta, line 15:
```suggestion
        [Obsolete("This method is deprecated and will be removed in the next major release. Use WorkFactor instead.")]
```

src/Okta.Sdk/Generated/UserFactorResolver.Generated.cs
```diff
@@ -70,6 +65,11 @@ protected override Type GetResolvedTypeInternal(IDictionary<string, object> data
                 return typeof(HardwareUserFactor);
             }
             
+            if (value.Equals("token:hotp"))
+            {
+                return typeof(CustomHotpUserFactor);
+            }
+
```
bryanapellanes-okta, line 72: We should consider updating the code generation templates to use a `Dictionary<string, Type>` instead of these `if` statements; we would get a micro performance improvement.

bryanapellanes-okta (approved):
One comment one suggestion; non-blocking.  👍

andriizhegurov-okta (approved):
Looks good

src/Okta.Sdk/PasswordCredentialHash.cs
laura-rodriguez: Thank you!

PR #454, opened 2020-11-16
Added Visual Studio compatibility information.

### LINKED ISSUE #446: Install fails via NuGet in VS 2015 targeting 4.7
## Current behavior

Picture1
Picture2

When installing Okta.Sdk via NuGet in Visual Studio 2015 targeting .NET 4.7.2, the install fails with the error:

"Could not install package 'Okta.Sdk 3.2.0'. You are trying to install this package into a project that targets .NETFramework, Version=v4.7.2, but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author."

Upgrading Visual Studio to a newer version resolves the issue. 

This was reproduced on multiple machines.

## Expected behavior

This may be behaving as expected if VS 2015 is no longer supported but if that is the case, it should be noted in the read-me. 

## Minimal reproduction of the problem with instructions

To reproduce, install VS 2015 (any build) in Windows, use NuGet to search for Okta.Sdk, and install. 

## Environment
 
- OS: Windows 10
- .Net Framework: 
4.7.2

andriizhegurov-okta:
## Issue \#

Fix #446 
https://oktainc.atlassian.net/browse/OKTA-340984

## Code

- [ ] Unit test(s)
- [ ] Implementation

## Additional Context

Added Visual Studio compatibility information to README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ The SDK is compatible with:
 * .NET Framework 4.6.1 or higher
 * .NET Core 3.0 or higher
 
+Visual Studio 2017 or newer is required as previous versions are not compatible with the above frameworks.
+
 ### Install using Nuget Package Manager
  1. Right-click on your project in the Solution Explorer and choose **Manage Nuget Packages...**
  2. Search for Okta. Install the `Okta.Sdk` package.


bryanapellanes-okta (approved):
👍

laura-rodriguez (approved):

PR #463, opened 2020-12-21
OKTA-337528 .NET Sdk Update readme with warning...

### LINKED ISSUE #442: Suggestion for the readme
This statement is the suggested way of "Find the desired user" in several places in the readme:
`var user = await client.Users.FirstOrDefaultAsync(x => x...`

We found out, the hard way, that this might not be a good idea when the collection of Users is large.

May we suggest that Users.FirstOrDefaultAsync is replaced with another method to "Find the desired user" in the readme?
Or perhaps clearify the side effects (with a warning label) of the method, when amount of Users is large?

Thanks!

andriizhegurov-okta:
about using await client.Users.FirstOrDefaultAsync(x => x... with many users

## Issue \#

Fix #442 
https://oktainc.atlassian.net/browse/OKTA-337528

## Code

- [ ] Unit test(s)
- [ ] Implementation

## Additional Context

In code samples `GetUserAsync` is suggested instead of `FirstOrDefaultAsync` to find a user

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,8 +229,8 @@ await vader.DeactivateOrDeleteAsync();
 ### List a User's Groups
 
 ``` csharp
-// Find the desired user
-var user = await client.Users.FirstOrDefaultAsync(x => x.Profile.Email == "darth.vader@imperial-senate.gov");
+// Retrieve the desired user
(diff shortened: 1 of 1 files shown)


bryanapellanes-okta (approved):
This looks good to me.

PR #480, opened 2021-03-24
Allow XML responses

andriizhegurov-okta:
## Code

- [ ] Unit test(s)
- [x] Implementation

## Current behavior

The SDK doesn't allow XML responses 

## Additional Context

Example of working with an XML response:
```csharp
    var appMetadata = await client.Applications.AppSamlMetadataAsync(app.Id);
    var xmlString = appMetadata["_rawData"].ToString();

    XmlDocument doc = new XmlDocument();
    doc.LoadXml(xmlString);
//..................
```

### DIFF
--- a/src/Okta.Sdk/HttpResponse{T}.cs
+++ b/src/Okta.Sdk/HttpResponse{T}.cs
@@ -4,6 +4,7 @@
 // </copyright>
 
 using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


bryanapellanes-okta (commented):
I think we should review the payload handlers implementation to see if a pattern is already established for handling this.

PR #483, opened 2021-03-25
Update migrating guide and readme.

laura-rodriguez:
## Code

- [ ] Unit test(s)
- [ ] Implementation

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -26,7 +26,7 @@ _Now:_
 
 ```
 
-Since this is a breaking change in the default behavior, Okta.Sdk was published with version numbers starting from 5.0.0.
+Since this is a breaking change in the default behavior, the major version of Okta.Sdk was incremented to 5; the latest version is now 5.0.0.
 
 If you were relying on this behavior, make sure to update your code and verify the resource is not `null` before accessing its properties.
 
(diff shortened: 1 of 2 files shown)


andriizhegurov-okta (approved):
👍

MIGRATING.md
```diff
@@ -2,6 +2,34 @@
 
 This library uses semantic versioning and follows Okta's [library version policy](https://developer.okta.com/code/library-versions/). In short, we don't make breaking changes unless the major version changes!
 
+## Migrating from 4.x to 5.x
+
+In previous versions, null resource properties would result in a resource object with all its properties set to `null`. Now, null resource properties would result in `null` property value.
```
bryanapellanes-okta, line 7:
```suggestion
In previous versions, null resource properties would result in a resource object with all its properties set to `null`. Now, null resource properties will result in `null` property value.
```

MIGRATING.md
```diff
@@ -2,6 +2,34 @@
...
+
+```
+{                                                 deserializedResource.Prop1.Should().Be("Hello World!");          
+    prop1 : "Hello World!",         =>            deserializedResource.NestedObject.Should().BeNull();
+    nestedObject: null                            
+}
+
+```
+
+Since this is a breaking change in the default behavior, Okta.Sdk was published with version numbers starting from 5.0.0.
```
bryanapellanes-okta, line 29:
```suggestion
Since this is a breaking change in the default behavior, the major version of Okta.Sdk was incremented to 5; the latest version is now 5.0.0.
```

bryanapellanes-okta (approved):
Some minor suggestions.

PR #491, opened 2021-07-01
OKTA-387100  .NET Mgmt SDK - Document how to set the "kid" value when creating the Client

andriizhegurov-okta:
Added example of initialization of RSA key for `OktaClientConfiguration`

Ticket: **OKTA-387100 .NET Mgmt SDK - Document how to set the "kid" value when creating the Client**
While our Management SDKs indicate how you can set up your YAML config to include the "kid" value associated with the private key you are using to instantiate your client (when using OAuth serice ap), https://github.com/okta/okta-sdk-dotnet#yaml-configuration, there is nothing in the documentation to explain how to do this inline (such as via the OktaClientConfiguration)

See existing README coverage here: https://github.com/okta/okta-sdk-dotnet#oauth-20

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,33 @@ var client = new OktaClient(new OktaClientConfiguration
 });
 ```
 
+Key object for assigning to the PrivateKey can be created and initialized inline like in this example for RSA key:
+
+``` csharp
+var privateKey = new JsonWebKeyConfiguration
+{
+    P = "{{P}}",
+    Kty = "RSA",
+    Q = "{{Q}}",
+    D = "{{D}}",
+    E = "{{E}}",
+    Kid = "{{P}}",
+    Qi = "{{Qi}}"
+};
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,6 +138,33 @@ var client = new OktaClient(new OktaClientConfiguration
 });
 ```
 
+Key object for assigning to the PrivateKey can be created and initialized inline like in this example for RSA key:
```
bryanapellanes-okta, line 141: Can we add more context as to what this is for?  Maybe include a link to more information about why someone would want to do this?

bryanapellanes-okta (commented):
Looking for more context.

README.md
andriizhegurov-okta: I've added the all the info I have to the PR's description

bryanapellanes-okta (approved):

PR #499, opened 2021-08-26
OKTA-414858 .NET SDK > Add support for new APIs that come in openAPI …

andriizhegurov-okta:
…spec 2.7.0

## Code

- [x] Unit test(s)
- [x] Implementation

## Additional Context

Regenerate .Net SDK with the latest OpenAPI (2.6.0 actually, not 2.7.0)
Add Domains API and Zones API

### DIFF
--- a/src/Okta.Sdk.IntegrationTests/DomainScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTests/DomainScenarios.cs
@@ -1,0 +1,63 @@
+﻿using System;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Xunit;
+
+namespace Okta.Sdk.IntegrationTests
+{
+    public class DomainScenarios
+    {
+        [Fact]
+        public async Task CreateDomain()
+        {
+            var oktaClient = TestClient.Create();
+            var domainsClient = oktaClient.Domains;
+            var domainName = $"domain{new Random().Next(1000000)}.example.com";
+            var newDomain = new Domain
+            {
+                DomainName = domainName,
+                CertificateSourceType = DomainCertificateSourceType.Manual,
+            };
+
+            var createDomainResult = await domainsClient.CreateDomainAsync(newDomain);
+            createDomainResult.DomainName.Should().Be(domainName);
+            createDomainResult.CertificateSourceType.Should().Be(DomainCertificateSourceType.Manual);
+            createDomainResult.ValidationStatus.Should().Be(DomainValidationStatus.NotStarted);
+            createDomainResult.Id.Should().NotBeNullOrEmpty();
+
+            await domainsClient.DeleteDomainAsync(createDomainResult.Id);
+        }
+
+        [Fact]
+        public async Task GetDomain()
+        {
+            var oktaClient = TestClient.Create();
+            var domainsClient = oktaClient.Domains;
+            var domainName = $"domain{new Random().Next(1000000)}.example.com";
+            var newDomain = new Domain
+            {
+                DomainName = domainName,
+                CertificateSourceType = DomainCertificateSourceType.Manual,
+            };
+
+            var createDomainResult = await domainsClient.CreateDomainAsync(newDomain);
+
+            // Find the domain among all domains
+            var domainList = await domainsClient.ListDomainsAsync();
+            domainList.Domains.Should().Contain(x => x.DomainName == domainName);
+
+            // Get the domain by Id
+            var getDomainResult = await domainsClient.GetDomainAsync(createDomainResult.Id);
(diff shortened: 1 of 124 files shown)


src/Okta.Sdk.IntegrationTests/DomainScenarios.cs
```diff
@@ -0,0 +1,63 @@
...
+            var oktaClient = TestClient.Create();
+            var domainsClient = oktaClient.Domains;
+            var domainName = $"domain{new Random().Next(1000000)}.example.com";
+            var newDomain = new Domain
+            {
+                DomainName = domainName,
+                CertificateSourceType = DomainCertificateSourceType.Manual,
+            };
+
+            var createDomainResult = await domainsClient.CreateDomainAsync(newDomain);
```
bryanapellanes-okta, line 22: Is the use of `var` the established standard?  It would be nice to know what this type actually is by looking at the code without having to look at the definition of CreateDomainAsync.

src/Okta.Sdk.IntegrationTests/NetworkZoneScenarios.cs
```diff
@@ -0,0 +1,134 @@
...
+                    Value = "1.1.1.1/24",
+                });
+
+                var updateZoneResponse = await zonesClient.UpdateNetworkZoneAsync(createZoneResponse, createZoneResponse.Id);
+                updateZoneResponse.Name.Should().Be(createZoneResponse.Name);
+                updateZoneResponse.Gateways.Count.Should().Be(3);
+            }
+            finally
+            {
+                await zonesClient.DeleteNetworkZoneAsync(createZoneResponse.Id);
```
bryanapellanes-okta, line 45: I'm not personally a fan of testing multiple different operations in a single unit test.  If any of the operations create, update or delete fail it will not be immediately obvious which operation needs attention.  Is it possible to specify failure messages that expose which operation failed if any?

src/Okta.Sdk.IntegrationTests/NetworkZoneScenarios.cs
```diff
@@ -0,0 +1,134 @@
...
+            NetworkZone newZone = BuildNetworkZoneObject();
+            var createZoneResponse = await zonesClient.CreateNetworkZoneAsync(newZone);
+            try
+            {
+                var allNetworkZones = zonesClient.ListNetworkZones();
+                var createdZone = await allNetworkZones.FirstOrDefaultAsync(z => z.Id == createZoneResponse.Id);
+
+                createdZone.Should().NotBeNull();
+
+                var getResponse = await zonesClient.GetNetworkZoneAsync(createZoneResponse.Id);
```
bryanapellanes-okta, line 83: I think this should be two different tests.  One for `List` and one for `Get`, for reasons previously stated.

bryanapellanes-okta (commented):
Looks fine to me; I'm interested in your thoughts on my comments before approving.

src/Okta.Sdk.IntegrationTests/NetworkZoneScenarios.cs
andriizhegurov-okta: Agree, I'll split it into two.

src/Okta.Sdk.IntegrationTests/NetworkZoneScenarios.cs
andriizhegurov-okta: It's a cleanup code that cannot be separated from the other. I need this `delete` to be called after `create` in this order.

src/Okta.Sdk.IntegrationTests/DomainScenarios.cs
andriizhegurov-okta: This makes whole code cleaner, I think using `var` is acceptable for small internal functions like these ITs.

src/Okta.Sdk/Generated/UserSchemaAttributeMaster.Generated.cs
```diff
@@ -17,9 +17,16 @@ namespace Okta.Sdk
...
-        public string Type 
+        public IList<IUserSchemaAttributeMasterPriority> Priority 
         {
-            get => GetStringProperty("type");
+            get => GetArrayProperty<IUserSchemaAttributeMasterPriority>("priority");
+            set => this["priority"] = value;
+        }
+        
+        /// <inheritdoc/>
+        public UserSchemaAttributeMasterType Type
```
laura-rodriguez, line 27: Let's add details about this change, explaining why this is not technically a breaking change. Let's add a small UT.
string profile.Type;

PR #506, opened 2021-09-28
Add sleep to wait for job to be finished on GroupSchema and UserSchema ITs

laura-rodriguez:
Fix OKTA-431664.


bryanapellanes-okta (approved):
👍

PR #514, opened 2021-10-29
OKTA-441966: add github workflow file (github action) to regenerate and test the latest @okta/openapi

bryanapellanes-okta:
- added `.github/workflows/regenerate-and-test.yml` which defines steps to regenerate and test the Sdk from the latest version of the `okta/openapi` spec: see example run here: https://github.com/okta/okta-sdk-dotnet/runs/4048205948?check_suite_focus=true


laura-rodriguez (approved):
👍

PR #523, opened 2021-11-19
fix handling of null fields on put requests in .net management sdk

andriizhegurov-okta:
## Issue \#

OKTA-253671 Fix handling of null fields on PUT requests in .NET management SDK

## Code

- [ ] Unit test(s)
- [x] Implementation

## Additional Context

Allow null values in payload for this endpoint https://developer.okta.com/docs/reference/api/users/#update-user


src/Okta.Sdk/Generated/User.Generated.cs
```diff
@@ -14,7 +14,9 @@
 namespace Okta.Sdk
 {
     /// <inheritdoc/>
+    [ResourceObject(NullValueHandling = ResourceNullValueHandling.Include)] 
     public sealed partial class User : Resource, IUser
+
```
bryanapellanes-okta, line 19: Did the template change to add a space here?

openapi/errata.js
```diff
@@ -146,6 +146,8 @@ const modelErrata = [
   { path: 'CSRMetadataSubjectAltNames', rename: 'CsrMetadataSubjectAltNames', renameReason: 'Pattern consistency' },
   { path: 'UserSchema', includeNullValues: true },
   { path: 'GroupSchema', includeNullValues: true },
+  { path: 'User', includeNullValues: true },
+
```
bryanapellanes-okta, line 150: Is the extra line intentional?

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 Running changelog of releases since `3.1.1`
 
 
+## v5.3.1
+
+- Add support for null values for User object. Now it's possible to remove for example the second email by setting it to null like this:
```
bryanapellanes-okta, line 7:
```suggestion
- Add support for null User object property values. Now it's possible to remove properties, for example, the second email, by setting it to null like the following:
```

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
...
+
+```csharp
+    userProfile["secondEmail"] = null;
+```
+
+This will be included in a request payload. Previously, all null values were stripped from the payload.
```
bryanapellanes-okta, line 13:
```suggestion
This will be included in the request payload. Previously, all null values were stripped from the payload. 
```

bryanapellanes-okta (commented):
A couple questions and comments.

bryanapellanes-okta:
There appear to be some test failures.

src/Okta.Sdk/Generated/User.Generated.cs
andriizhegurov-okta: yes, may be we need to check the template. The same line is added for example in `GroupSchema.Generated.cs` which is also using the the same errata rule

openapi/errata.js
andriizhegurov-okta: no, I will remove it, thanks

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 Running changelog of releases since `3.1.1`
 
 
+## v5.3.1
```
laura-rodriguez, line 5: This will be a breaking change, so a bump in the major version is required.

src/Okta.Sdk/Generated/User.Generated.cs
laura-rodriguez: I believe the  [ResourceObject(NullValueHandling = ResourceNullValueHandling.Include)]  attribute should applied to UserProfile instead of user.

andriizhegurov-okta: not really, we're talking about this endpoint https://developer.okta.com/docs/reference/api/users/#update-user
Setting the attribute for `User` will affect UserProfile as well, but setting it for UserProfile will not have effect when updating User object

PR #559, opened 2022-06-27
OKTA-506941: add ITs for user api

bryanapellanes-okta:


oas3/output-csharp-netcore/src/Okta.Sdk.IntegrationTest/UserScenarios.cs
```diff
@@ -0,0 +1,1313 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Okta.Sdk.IntegrationTest
+{
+    public class UserScenarios
+    {
+        UserApi _userApi;
```
laura-rodriguez, line 17: Nit: Make them explicitely private

laura-rodriguez (approved):

PR #573, opened 2022-08-02
OKTA-498007: regenerate sdk from open api spec v2.13.0

bryanapellanes-okta:
This PR constitutes the result of updating the okta open api spec to v2.13.0 and running `npm run generate`

### DIFF
--- a/src/Okta.Sdk/Generated/IGroupsClient.Generated.cs
+++ b/src/Okta.Sdk/Generated/IGroupsClient.Generated.cs
@@ -16,18 +16,6 @@ namespace Okta.Sdk
     public partial interface IGroupsClient
     {
         /// <summary>
-        /// Enumerates groups in your organization with pagination. A subset of groups can be returned that match a supported filter expression or query.
-        /// </summary>
-        /// <param name="q">Searches the name property of groups for matching value</param>
-        /// <param name="filter">Filter expression for groups</param>
-        /// <param name="after">Specifies the pagination cursor for the next page of groups</param>
-        /// <param name="limit">Specifies the number of group results in a page</param>
-        /// <param name="expand">If specified, it causes additional metadata to be included in the response.</param>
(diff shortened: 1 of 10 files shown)


src/Okta.Sdk/Generated/IGroupsClient.Generated.cs
```diff
@@ -19,12 +19,13 @@ public partial interface IGroupsClient
...
         /// <param name="q">Searches the name property of groups for matching value</param>
-        /// <param name="search">Filter expression for groups</param>
+        /// <param name="filter">Filter expression for groups</param>
         /// <param name="after">Specifies the pagination cursor for the next page of groups</param>
         /// <param name="limit">Specifies the number of group results in a page</param>
         /// <param name="expand">If specified, it causes additional metadata to be included in the response.</param>
+        /// <param name="search">Searches for groups with a supported filtering expression for all attributes except for _embedded, _links, and objectClass</param>
         /// <returns>A collection of <see cref="IGroup"/> that can be enumerated asynchronously.</returns>
-        ICollectionClient<IGroup> ListGroups(string q = null, string search = null, string after = null, int? limit = 10000, string expand = null);
+        ICollectionClient<IGroup> ListGroups(string q = null, string filter = null, string after = null, int? limit = 10000, string expand = null, string search = null);
```
laura-rodriguez, line 28: We should provide a manual overload marked as deprecated with the previous interface to avoid breaking changes.

laura-rodriguez (approved):

src/Okta.Sdk/Generated/IGroupsClient.Generated.cs
bryanapellanes-okta: Local testing reveals this would cause an ambiguous call compile time error in [GroupsClient](https://github.com/okta/okta-sdk-dotnet/blob/master/src/Okta.Sdk/GroupsClient.cs#L37)

laura-rodriguez (approved):

PR #577, opened 2022-08-09
Allow additionalProperties by default, and test User Profile.

laura-rodriguez:
- Add support for additional Properties.
- Set "disallowAdditionalPropertiesIfNotPresent": false
- Update ITs to test dynamic props in UserProfile
- Fix OKTA-518586

### DIFF
--- a/src/Okta.Sdk.IntegrationTest/UserScenarios.cs
+++ b/src/Okta.Sdk.IntegrationTest/UserScenarios.cs
@@ -303,6 +303,8 @@ namespace Okta.Sdk.IntegrationTest
             {
                 // Update profile
                 createdUser.Profile.NickName = nickName;
+                createdUser.Profile.AdditionalProperties = new Dictionary<string, object>();
+                createdUser.Profile.AdditionalProperties["homeworld"] = "Planet Earth";
                 var updateUserRequest = new UpdateUserRequest
                 {
                     Profile = createdUser.Profile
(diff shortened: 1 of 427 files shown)


src/Okta.Sdk.IntegrationTest/UserScenarios.cs
```diff
@@ -303,16 +303,20 @@ public async Task UpdateUserProfile(string nickName)
             {
                 // Update profile
                 createdUser.Profile.NickName = nickName;
+                createdUser.Profile.AdditionalProperties = new Dictionary<string, object>();
+                createdUser.Profile.AdditionalProperties["homeworld"] = "Planet Earth";
```
arvindkrishnakumar-okta, line 307: This requires that the property `homeworld` is first added to the user profile schema in an Org via the console. This means that this IT is bound to fail if run in a new Org that does not have this item added.

arvindkrishnakumar-okta: Ok, I think in Java SDK it is not a good idea to do this as it is run in PDVs across multiple Orgs and will likely throw failures.

laura-rodriguez: Yeah, that makes sense.

arvindkrishnakumar-okta (approved):
LGTM

PR #590, opened 2022-10-03
Enable brand API tests.

laura-rodriguez:
The underlying API issue was fixed and we can enable back disabled ITs.
Fix OKTA-489495


bryanapellanes-okta (approved):
:shipit:

Issue #589, opened 2022-10-03
Deactivate IDP does not deactivate the routing rule pointing to the IDP unlike direct API call or the Okta UI


aqib-bhat:
### Describe the bug?

- Using Okta dotnet SDK and calling `OktaClientObject.IdentityProviders.DeactivateIdentityProviderAsync(someOktaIdpId, cancellationToken);` results in the IDP getting deactivated but a routing rule made for it remains Active.
- This is unlike what happens in the Okta UI and through direct calls to the Okta API.
  - On deactivating an Active IDP, its routing rule is also deactivated.
- I also tried the `IOktaClient`.`PostAsync<IdentityProvider>` but that also behaved the same way.

---

**P.S.** On a side note: The Okta UI has also a weird bug where on trying to deactivate an IDP, the warning says that the following routing rules will be disabled but some of them are not pointing to the IDP being disabled but just happen to match part of the name. On clicking OK though, only the actually related routing rule gets deactivated but the message is confusing.
![image](https://user-images.githubusercontent.com/14251134/193491559-21656db9-9b3d-4930-b835-17eb0f9b49dc.png)

### What is expected to happen?

Just like what happens in the UI and through direct calls to the Okta API, deactivating an IDP through the Okta dotnet SDK should result in the IDP's routing rule(s) also getting deactivated.

### What is the actual behavior?

Unlike what happens in the UI and through direct calls to the Okta API, deactivating an IDP through the Okta dotnet SDK only results in the IDP getting deactivated while it's routing rule(s) remain(s) Active.

### Reproduction Steps?

- Create a SAML2 IDP in Okta.
- Add a routing rule in the Default IDP_DISCOVERY policy, which directs to the above created IDP when the email of the user ends in some given email domains.
  - The IDP and routing rule will be Active at this stage.
- **In the Okta UI**, deactivate the IDP. Accept / Click OK for the warning rule that the associated routing rules will also be deactivated.
  - Check that the IDP and its routing rule both are now Inactive.
  - Make the IDP and its routing rule both Active again.
- **Hit the Okta API directly** to deactivate the IDP.
  - In the Okta UI,  check that the IDP and its routing rule both are now Inactive.
  - Make the IDP and its routing rule both Active again.
- **Use Okta dotnet SDK** to deactivate the IDP (`OktaClientObject.IdentityProviders.DeactivateIdentityProviderAsync(someOktaIdpId, cancellationToken);`).
  - In the Okta UI,  you will find that:
[...]

PR #598, opened 2022-10-28
Add IT to verify issue #589

laura-rodriguez:
Add IT to verify that deactivating an IDP also deactivates its routing rules.
Related to #589


bryanapellanes-okta (approved):
LGTM 🚢

PR #617, opened 2023-02-03
Fix "[.NET OASv3] Listing IAM roles doesn't work"

### LINKED ISSUE #607: Listing roles does not work
### Describe the bug?

`IamRole` object seems to require a `Permissions` property
image

but this property does not exist in the returned payload from `ListRolesAsync` 

image

which is causing an exception

image

Listing roles properly

Throwing exception

### Reproduction Steps?

1.  Create `RoleApi`
2.  Call `ListRolesAsync`

### Additional Information?

none

### .NET Version

net6.0

### SDK Version

6.0.1

### OS version

windows 11

laura-rodriguez:
- Update OpenApi templates to make interfaces partial
- Add partial RoleApi class with obsolete methods
- Fix spec
- Regenerate solution with updated templates and spec
Fix #607


src/Okta.Sdk/Api/RoleApi.cs
```diff
@@ -62,7 +62,7 @@ public interface IRoleApiAsync : IApiAccessor
         /// <param name="instance"></param>
         /// <param name="cancellationToken">Cancellation Token to cancel the request.</param>
         /// <returns>Task of IamRole</returns>
-        System.Threading.Tasks.Task<IamRole> CreateRoleAsync(  IamRole instance , System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
+        System.Threading.Tasks.Task<IamRole> CreateRoleAsync(  CreateIamRoleRequest instance , System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
```
bryanapellanes-okta, line 65: nitpick: I feel like it would be preferred for parameters that are some type of request object be named `request`.

bryanapellanes-okta (approved):
minor comment, but no blocking issues 👍

src/Okta.Sdk/Api/RoleApi.cs
laura-rodriguez: I followed the same convention as existing APIs for consistency.

PR #620, opened 2023-02-23
Renew token when response is 401 (privateKey)

### LINKED ISSUE #616: SDK calls with private key authorization mode start returning empty results after inner access token expired
### Describe the bug?

SDK calls with private key authorization mode start sending empty results once inner access token that is generating by SDK expired.

After inner access token expired it should be refreshed automatically. According to sources it is supporting but seems that there are some issue with that functionality (maybe with access token assigning inside header to call okta rest api internally).

After inner access token expired SDK calls stat returning empty results without any exception.

### Reproduction Steps?

 1. Create API Services okta application.
 2. Generate Private Key for new application.
 3. Grant okta.groups.read scope.
 4. Create new .NET 7.0 project with Okta.Sdk dependency and use following code to get okta groups
 
 ```
var configuration = new Configuration
{
	AuthorizationMode = AuthorizationMode.PrivateKey,
	OktaDomain = <okta domain>,
	ClientId = <okta application client id>,
	PrivateKey = new JsonWebKeyConfiguration(<okta application private key>),
	Scopes = new HashSet<string> {"okta.groups.read"}
});

var groupsApi = new GroupApi(configuration);
var groups1 = await groupApi.ListGroups().ToListAsync(); // groups1 contains valid okta groups
<wait while inner access token expired, for example 2 hours>
var groups2 = await groupApi.ListGroups().ToListAsync(); // groups2 list is empty and there are no errors 
```

### Additional Information?

_No response_

### .NET Version

7.0.101

### SDK Version

6.0.1

### OS version

_No response_

laura-rodriguez:
- Increase token retry count from 1 to 2
- Update Polly retry to use onRetryAsync instead of onRetry
- Clean Authorization headers before updating the token to avoid duplicated entries
- Pass configuration when retrieving the next page
- Update templates
- Add IT to test retry
- Fix #616

### DIFF
--- a/openapi3/templates/ApiClient.mustache
+++ b/openapi3/templates/ApiClient.mustache
@@ -587,11 +587,11 @@ namespace {{packageName}}.Client
             return result;
         }
         
-        private Task<IRestResponse> ExecuteAsyncWithRetryHeaders(Context context, IRestRequest request, RestClient client)
+        private async Task<IRestResponse> ExecuteAsyncWithRetryHeaders(Context context, IRestRequest request, RestClient client)
         {
             DefaultRetryStrategy.AddRetryHeaders(context, request);
             DefaultOAuthTokenProvider.AddOrUpdateAuthorizationHeader(context, request);
-            return client.ExecuteAsync(request);
+            return await client.ExecuteAsync(request);
         }
 
         #region IAsynchronousClient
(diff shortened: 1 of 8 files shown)


bryanapellanes-okta (approved):
LGTM 👍

PR #623, opened 2023-03-15
Fix "UpdateProfileMapping" doesn't update properties"

### LINKED ISSUE #618: UpdateProfileMappingAsync doesn't update properties
### Describe the bug?

Hi. I'm trying to add a property mapping to the user profile mapping, but my changes are ignored.

When calling `UpdateProfileMappingAsync` the profile mapping should be updated in okta, and the response object should contain the updates

Nothing changes

### Reproduction Steps?

Create a SAML Identity Provider and find its profile mapping id (I believe it will happen with any profile mapping for any application or Identity Provider but I didn't test it).
```csharp
var profileMappingApi = new ProfileMappingApi();
var mapping = await profileMappingApi.GetProfileMappingAsync("your-mapping-id");

mapping.Properties.Add("userType", new ProfileMappingProperty
{
    PushStatus = ProfileMappingPropertyPushStatus.PUSH,
    Expression = "appuser.firstName" // just for example
});

var updatedMapping = await profileMappingApi.UpdateProfileMappingAsync("your-mapping-id", fullMapping);
```

### Additional Information?

The same scenario works in Postman.

### .NET Version

7.0.101

### SDK Version

6.0.2

### OS version

19045        Microsoft Windows 10 Pro  64-bit          10.0.19045

laura-rodriguez:
- Update spec to make ProfileMapping.Properties editable.
- Add ITs
- Fix #618


bryanapellanes-okta (approved):
I see no issues 👍

PR #627, opened 2023-03-21
Fix 575483 - update user async should do full replace

laura-rodriguez:


openapi3/management.yaml
```diff
@@ -13930,7 +13930,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/UpdateUserRequest'
+              $ref: '#/components/schemas/User'
```
bryanapellanes-okta, line 13933: Will this require broader review, or is this somehow specific to just the .NET Sdk? Or, has it already been reviewed?

bryanapellanes-okta (commented):
Will this require a version update?

laura-rodriguez:
> Will this set of changes require a version update?

I kept the previous interface and marked it as obsolete to avoid breaking changes.

bryanapellanes-okta (approved):
👍

PR #635, opened 2023-04-11
Throw a proper timeout exception when the client times out.

### LINKED ISSUE #632: API Calls Trap Request Timeout Exceptions
### Describe the bug?

When making an API call, if the request times out, there is no exception propagated to the caller. Instead a default value, depending on the method's signature, is returned. For example, `UserApi.GetUserAsync(string)` will return `null`. `UserApi.ListUsers()` will return an empty list once you enumerate it.

I am not sure if this is a bug or by design so I would appreciate a little bit of clarity. An API call that returns `null` can possibly be used to detect a connection timeout but something like an empty list makes this very difficult to detect. I have a scenario where I need to tell the difference between the API not finding any data and the API actually failing because of a transient connection error.

Thanks in advance.

Any API call, such as  `UserApi.GetUserAsync(string)` , should throw an exception if there is a connection timeout. Not trap it.

The exception is trapped.

### Reproduction Steps?

1. Create a `UserApi` instance. Use a dummy proxy (i.e. a non existent proxy address) to simulate a connection timeout
2. Call `UserApi.ListUsers()` and enumerate the list
3. An empty list is returned because of the connection timeout

### Additional Information?

_No response_

### .NET Version

7.0.201

### SDK Version

6.0.7

### OS version

_No response_

laura-rodriguez:
## Summary

RestSharp doesn't throw an exception when the client times out. Instead, it returns a response with `Status = Timedout`. In this PR, I check for that Status property and throw a proper TimeoutException.

Fixes #632 

## Type of PR

- [x] Bug Fix (non-breaking fixes to existing functionality)
- [ ] New Feature (non-breaking changes that add new functionality)
- [ ] Documentation update
- [x] Test Updates
- [ ] Other (Please describe the type)

## Test Information

- [x] My PR required test updates


src/Okta.Sdk/Client/Configuration.cs
```diff
@@ -728,7 +758,7 @@ public static string ToDebugReport()
             report += "    OS: " + System.Environment.OSVersion + "\n";
             report += "    .NET Framework Version: " + System.Environment.Version  + "\n";
             report += "    Version of the API: 3.0.0\n";
-            report += "    SDK Package Version: 6.0.7\n";
+            report += "    SDK Package Version: 6.0.8\n";
```
bryanapellanes-okta, line 761: The version should probably be retrieved from the assembly to make sure we don't accidently report the wrong version if this line is not updated in the future, see [Version Class](https://learn.microsoft.com/en-us/dotnet/api/system.version?view=net-8.0#remarks):

```
Version version = Assembly.GetExecutingAssembly().GetName().Version;
string version  = $"{version.Major}.{version.Minor}.{version.Build}";
```

bryanapellanes-okta (approved):
Non blocking suggestion. :shipit:

src/Okta.Sdk/Client/Configuration.cs
laura-rodriguez: This is all taken care of by the codegen, you update the config.json, and the generator does the work for us.

PR #636, opened 2023-04-14
Add missing UserFactors' resend endpoint.

### LINKED ISSUE #633: Resend SMS as part of enrollment

When enrolling the SMS factor, a user may want a new code sent to the same phone.  The Factor API supports resending a code via SMS during enrollment of the factor.  The SDK does not include this feature.

[UserFactorApi.cs](https://github.com/okta/okta-sdk-dotnet/blob/c23340e9da793532e17cf93bf981c4fd2ed07a42/src/Okta.Sdk/Api/UserFactorApi.cs)

https://developer.okta.com/docs/reference/api/factors/#resend-sms-as-part-of-enrollment

### Additional Information?

_No response_

laura-rodriguez:
Fixes #633


docs/NetworkZoneApi.md
```diff
@@ -86,9 +86,9 @@ Name | Type | Description  | Notes
 | Status code | Description | Response headers |
 |-------------|-------------|------------------|
 | **200** | Success |  -  |
-| **403** | Forbidden |  -  |
-| **404** | Not Found |  -  |
-| **429** | Too Many Requests |  -  |
```
bryanapellanes-okta, line 91: Was it intentional that the descriptions were removed?  It seems preferable to keep them.

laura-rodriguez: It all depends on the spec, the code is autogenerated.

bryanapellanes-okta: I thought that was the case; should we review the spec more closely?

openapi3/management.yaml
```diff
@@ -14582,6 +14582,57 @@ paths:
...
+        '404':
+          $ref: '#/components/responses/ErrorResourceNotFound404'
+        '429':
+          $ref: '#/components/responses/ErrorTooManyRequests429'
+      security:
+        - apiToken: []
+        - oauth2:
+            - okta.users.manage
+      tags:
+        - UserFactor
```
bryanapellanes-okta, line 14635: Do the changes shown here propagate down from changes to the spec, or do these changes need to be propagated back up to the spec for everyone else to consume?

bryanapellanes-okta (approved):
Thanks for the one on one explanation 👍

PR #647, opened 2023-05-18
Fix "Null response after hitting rate limit" issue 

### LINKED ISSUE #638: Create/Update Account returns NULL when okta tenant hits rate limits
### Describe the bug?

When okta tenant is under heavy load and rate limit events are raised, CreateUserAsync and PartialUpdateUserAsync on IUserApi returns NULL object.

If the request is unsuccessful, an exception is expected.  The SDK methods are defined to return an object, unless there is an exception condition preventing the operation from completing successfully. 

The SDK methods return null object, when the okta server is under heavy load -- ie, rate limit events reported for /api/v1/users (write and read).  The SDK does not throw an exception. The SDK hides the details of rate limit handling and retries, and it returns a null object if the operation is not successful. 

### Reproduction Steps?

Run a migration or test tool that floods the okta tenant with CreateAccountSync and PartialUpdateUserAsync requests that eventually triggers rate limit events on okta. Some of the requests will eventually begin to return null from these iUser methods. 

### Additional Information?

We can consistently trigger this behavior during a user migration from previous OAuth provider to Okta.  Because of this behavior and rate limiting, the migration tool has been modified to limit throughput that does not trigger rate limit events.  However, this increases the length of time required to migrate production user accounts into Okta. 

### .NET Version

mcr.microsoft.com/dotnet/sdk:6.0 

### SDK Version

6.0.8

### OS version

mcr.microsoft.com/dotnet/aspnet:6.0 runtime

laura-rodriguez:
- Fix #638 
- Update templates and regenerate solution
- Prepare release 6.0.11
- Add UTs


bryanapellanes-okta (approved):
👍

PR #648, opened 2023-06-20
Update templates to use the latest version of RestSharp

### LINKED ISSUE #606: Upgrade package to use RestSharp v107+

We are attempting to update our application from using okta’s .Net SDK v5x to v6.0.1.
In doing so, we run into versioning errors with the RestSharp library now used by the v6 SDK.
The SDK uses v106 of the RestSharp library. RestSharp introduced breaking changes in v107, which was released in January of 2022. Because the SDK uses v106, and we have other packages that use later versions of RestSharp (108) - packages which we can’t downgrade - we cannot use the v6 SDK.

Request that the okta .net sdk be updated to use a more recent version of the RestSharp library.

Would affect all endpoints in the SDK.

https://restsharp.dev/v107/#restsharp-v107

### Additional Information?

_No response_

laura-rodriguez:
## Summary

Upgrades RestSharp dependency to 110.2.0. 

Upgrading to RestSharp 110.2.0 requires a few breaking changes. Some interfaces, such as IRestRequest or IRestResponse, no longer exist. Also, the RestClient configuration has been moved to a new class called `RestClientOptions`. 
Several tests were updated to align with the new version.

Models have been updated due to a bug fix in GetHashCode. Now, StringEnum properties are checked for null values before calling GetHashCode().

Fixes #606 
 

## Type of PR

- [ ] Bug Fix (non-breaking fixes to existing functionality)
- [ ] New Feature (non-breaking changes that add new functionality)
- [ ] Documentation update
- [x] Test Updates
- [ ] Other (Please describe the type)

## Test Information

- [ ] My PR required test updates 

.NET Version:
Os Version:

## Signoff
- [ ] I have submitted a CLA for this PR
- [ ] Each commit message explains what the commit does
- [ ] I have updated documentation to explain what my PR does
- [ ] My code is covered by tests if required
- [ ] I checked StyleCop warnings on my code

### DIFF
--- a/openapi3/templates/ApiClient.mustache
+++ b/openapi3/templates/ApiClient.mustache
@@ -37,7 +37,7 @@ namespace {{packageName}}.Client
     /// <summary>
     /// Allows RestSharp to Serialize/Deserialize JSON using our custom logic, but only when ContentType is JSON.
     /// </summary>
-    internal class CustomJsonCodec : IRestSerializer, ISerializer,IDeserializer
+    internal class CustomJsonCodec : IRestSerializer, ISerializer, IDeserializer
     {
         private readonly IReadableConfiguration _configuration;
         private static readonly string _contentType = "application/json";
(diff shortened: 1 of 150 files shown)


openapi3/templates/ApiClient.mustache
```diff
@@ -499,18 +522,51 @@ namespace {{packageName}}.Client
                 client.ClientCertificates = configuration.ClientCertificates;
             }
 
-            return client;
+            return client;*/
```
bryanapellanes-okta, line 525: Is it desirable to have the code generator generate a section of commented code like this?

src/Okta.Sdk/Okta.Sdk.csproj
```diff
@@ -20,13 +20,13 @@
   </PropertyGroup>
```
bryanapellanes-okta, line 20: Should the version be update to 7.0?

openapi3/templates/ApiClient.mustache
laura-rodriguez: Good catch, I'll clean up.

bryanapellanes-okta (approved):
👍

PR #652, opened 2023-08-04
Update spec to v2023.07.0 and regenerate code

laura-rodriguez:
Review MIGRATING guide for details about changes: https://github.com/okta/okta-sdk-dotnet/pull/654


jcaldwell-tandemdiabetes:
@laura-rodriguez, the 6.x -> 7.x migration guide seems to be missing some changes and incorrect / inconsistent in some cases.  Was able to figure out the name changes -- that's not a problem, although several name changes on the groupsApi are missing from the guide.  The question I have is that the guide mentions several userApi methods are obsolete, replaced by other methods.  What's mentioned appears to be backwards -- ie, the obsoleted method remains, while the suggested new method does not exists.  For example:

[Obsolete("This method is obsolete. Use UpdateUserAsync(string userId, User user,...) instead.")]
System.Threading.Tasks.Task<User> UpdateUserAsync(string userId, UpdateUserRequest user, bool? strict = default(bool?), System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));

We see that UpdateUserAsync(string userId, User user,...)  is the obsoleted method.

laura-rodriguez:
Thanks for reporting this @jcaldwell-tandemdiabetes! 

There have been so many changes in the spec made by the API teams that it's possible some changes were missed in the migrating guide. Apologies for the inconvenience. Feel free to file a PR with the missing changes in the migrating guide if you want; no pressure at all, though 😃.

Regarding `UpdateUserAsync`, you're correct. You have to use ` UpdateUserAsync(string userId, UpdateUserRequest user, bool? strict = default(bool?)?`.  The API teams updated the spec, so this method now receives an `UpdateUserRequest` instead of a user.

PR #668, opened 2023-09-28
Update the collection templates to throw exceptions when token is expired

### LINKED ISSUE #658: Inconsistent Exception Handling
### Describe the bug?

Normally, one would expect methods that return only data from an http call, to throw when there are errors (Such as 401,403).  However, the function is returning an empty array in this case, which is misleading because that would indicate that no users were found rather than indicating a configuration problem.

Conversely when using a function that returns HTTP status and data, one would expect to get a response object, and not throw exceptions on errors, such as when calling ListUsersWithHttpInfoAsync.  However, the Okta SDK throws an exception in this case when the token is invalid.

The behaviors are completely flipped.   Effectively it is forcing me to use ListUsersWithHttpInfoAsync at the moment, because otherwise there would be no way to tell if the application is configured incorrectly.

ListUsersAsync function should throw errors when the call could not be made successfully.
ListUsersWithHttpInfoAsync should not throw exceptions for non-200 responses and instead should let the caller handle the error responses.

ListUsersAsync returns an empty array when an HTTP error occurs.
ListUsersWithHttpInfoAsync throws an error when the provided token is invalid.

### Reproduction Steps?

The description is sufficient.

### Additional Information?

_No response_

### .NET Version

7.0.306

### SDK Version

6.0.11

### OS version

Darwin MacBook-Pro-3.local 22.6.0 Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:05 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6000 arm64

laura-rodriguez:
Fix #658
Fix OKTA-637431

### DIFF
--- a/docs/UserType.md
+++ b/docs/UserType.md
@@ -9,7 +9,7 @@ Name | Type | Description | Notes
 **Default** | **bool** | A boolean value to indicate if this is the default User Type | [optional] [readonly] 
 **Description** | **string** | The human-readable description of the User Type | [optional] 
 **DisplayName** | **string** | The human-readable name of the User Type | 
-**Id** | **string** | The unique key for the User Type | [optional] [readonly] 
+**Id** | **string** | The unique key for the User Type | [optional] 
 **LastUpdated** | **DateTimeOffset** | A timestamp from when the User Type was most recently updated | [optional] [readonly] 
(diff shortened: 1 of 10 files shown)


bryanapellanes-okta (commented):
Some minor comments.

bryanapellanes-okta (approved):
👍

PR #671, opened 2023-10-03
Update the Links property of the VerifyUserFactorResponse model in the Open API spec to expose poll and cancel objects.

### LINKED ISSUE #665: Upgrade Okta.Sdk 6.0 to 7.0: MFA not able to get transactionId
### Describe the bug?

After I upgraded the Okta.Sdk from 6.0 to 7.0, I cannot fetch the transactionId anymore.
I have an app where I need to verify against Okta, where a user sent his username & password.
Then the Okta app on the user his phone gets a msg, where the user has to verify it is him.
With Okta.Sdk 6.0 I get a VerifyUserFactorResponse (at userFactorApi.VerifyFactorAsync) with some "links" from which i can retrieve the transactionId in the "Link" property.
It seems like the [https://github.com/okta/okta-sdk-dotnet/blob/master/src/Okta.Sdk/Model/VerifyUserFactorResponse.cs](https://github.com/okta/okta-sdk-dotnet/blob/master/src/Okta.Sdk/Model/VerifyUserFactorResponse.cs) doesn't corresponds with the answer from the api anymore (as mentioned in [https://developer.okta.com/docs/reference/api/factors/#response-example-25](https://developer.okta.com/docs/reference/api/factors/#response-example-25)

I have the following code
```c#
private async Task DoLogonAsync(string userName, string password, CancellationToken cancellationToken)
    {
        var authClient = new AuthenticationClient(new OktaClientConfiguration
        {
            OktaDomain = "someOktaDomain"
        });

        var authResponse = await authClient.AuthenticateAsync(new AuthenticateOptions
        {
            Username = userName,
            Password = password
        }, cancellationToken);


        if (authResponse.AuthenticationStatus == AuthenticationStatus.MfaRequired)
        {
            var oktaConfig = new Configuration
            {
                OktaDomain = "someOktaDomain",
                Token = "someOktaApiToken"
            };

            var userApi = new UserApi(oktaConfig);
            var user = await userApi.GetUserAsync(userName, cancellationToken);

            var userFactorApi = new UserFactorApi(oktaConfig);
            var factors = await userFactorApi.ListFactorsWithHttpInfoAsync(user.Id, cancellationToken);

            var factor = factors.Data.FirstOrDefault(x => x.FactorType == Okta.Sdk.Model.FactorType.Push);
            if (factor != null)
            {
                var result = await userFactorApi.VerifyFactorAsync(user.Id, factor.Id, null, cancellationToken: cancellationToken);

            }
        }
    }
```

Get a VerifyUserFactorResponse from which I can fetch the transactionId

The links in the VerifyUserFactorResponse are null.

### Reproduction Steps?

See code above

### Additional Information?

_No response_

### .NET Version

dotnet 7.0

### SDK Version

6.0 & 7.0

### OS version

_No response_

laura-rodriguez:
Fix OKTA-653501
Fix #665


src/Okta.Sdk/Model/VerifyUserFactorResponse.cs
```diff
@@ -79,7 +79,7 @@ public bool ShouldSerializeEmbedded()
         /// Gets or Sets Links
         /// </summary>
         [DataMember(Name = "_links", EmitDefaultValue = true)]
-        public LinksSelf Links { get; set; }
+        public VerifyUserFactorResponseLinks Links { get; set; }
```
bryanapellanes-okta, line 82: Is this a breaking change?

bryanapellanes-okta (commented):
Will this require a minor version bump?

src/Okta.Sdk/Model/VerifyUserFactorResponse.cs
3sRykaert: It is broken in current 7.0.0 branch, since the VerifyUserFactorResponse doesn't corresponds with the answer from the api.
So imho it's not a breaking change, but a bugfix.

3sRykaert (approved):

src/Okta.Sdk/Model/VerifyUserFactorResponse.cs
laura-rodriguez: This is one of those gray scenarios where you are technically making interface changes, but the original behavior never worked. In this case, I'm inclined to categorize this as a bug fix because:

* `VerifyUserFactorResponseLinks` is a superset of `LinksSelf`. `LinkSelf` exposed a Self property only, while `VerifyUserFactorResponseLinks` exposes `Self`, `Cancel` and `Poll`. So, if you had the following code line in your program, it won't break after this change:
```
var link = response.Links.Self.Href;
```
* As @3sRykaert mentioned, the response.Links property was unreliable because it didn't fully correspond with the server's response. 

Thoughts @bryanapellanes-okta ?

bryanapellanes-okta: > This is one of those gray scenarios where you are technically making interface changes, but the original behavior never worked. In this case, I'm inclined to categorize this as a bug fix because:
> 
> * `VerifyUserFactorResponseLinks` is a superset of `LinksSelf`. `LinkSelf` exposed a Self property only, while `VerifyUserFactorResponseLinks` exposes `Self`, `Cancel` and `Poll`. So, if you had the following code line in your program, it won't break after this change:
> 
> ```
> var link = response.Links.Self.Href;
> ```
> 
> * As @3sRykaert mentioned, the response.Links property was unreliable because it didn't fully correspond with the server's response.
> 
> Thoughts @bryanapellanes-okta ?

I suspected it was a scenario as described, 👍

bryanapellanes-okta (approved):
👍

PR #682, opened 2023-11-03
JTI Claim as a string instead of guid

sachinsatav:
## Summary
Latest Microsoft Identity Model package >7.0.0 and above no longer serializes Guid datatype ( see - https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f5fef2269a6eea5ba52d707d7235acf6a504db9f/src/Microsoft.IdentityModel.Tokens/Json/JsonSerializerPrimitives.cs#L860-L924) and instead throws an exception -

> System.InvalidOperationException: Something went wrong when creating the signed JWT. Verify your private key.
>  ---> System.ArgumentException: IDX11025: Cannot serialize object of type: 'System.Guid' into property: 'jti'.
>    at Microsoft.IdentityModel.Tokens.Json.JsonSerializerPrimitives.WriteObject(Utf8JsonWriter& writer, String key, Object obj)
>    at Microsoft.IdentityModel.Tokens.Json.JsonSerializerPrimitives.WriteObjects(Utf8JsonWriter& writer, IDictionary`2 dictionary)
>    at System.IdentityModel.Tokens.Jwt.JwtPayload.SerializeToJson()
>    at System.IdentityModel.Tokens.Jwt.JwtPayload.Base64UrlEncode()
>    at System.IdentityModel.Tokens.Jwt.JwtSecurityToken.get_EncodedPayload()
>    at System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.WriteToken(SecurityToken token)
>    at Okta.Sdk.Client.DefaultJwtGenerator.GenerateSignedJWT()
>    --- End of inner exception stack trace ---
>    at Okta.Sdk.Client.DefaultJwtGenerator.GenerateSignedJWT()
>    at Okta.Sdk.Api.OAuthApi.GetBearerTokenWithHttpInfoAsync(CancellationToken cancellationToken)
>    at Okta.Sdk.Api.OAuthApi.GetBearerTokenAsync(CancellationToken cancellationToken)
>    at Okta.Sdk.Client.DefaultOAuthTokenProvider.RequestAccessTokenAsync(CancellationToken cancellationToken)
>    at Okta.Sdk.Client.DefaultOAuthTokenProvider.GetAccessTokenAsync(Boolean forceRenew, CancellationToken cancellationToken)
>    at Okta.Sdk.Api.UserApi.GetUserWithHttpInfoAsync(String userId, CancellationToken cancellationToken)
[...]

PR #685, opened 2023-12-15
Introduce "JTI Claim as a string instead of guid" (#682) external contribution

laura-rodriguez:
This PR includes #682 and updates the codegen templates to port the solution into the codegen process.


bryanapellanes-okta (approved):
👍

PR #711, opened 2024-05-02
Add support for DPoP

laura-rodriguez:
* Add a DPoPProofJwtGenerator class responsbile for DPoP Proof JWT generation
* Update OAuthTokenProvider interface and default classes to return an OAuthTokenResponse instead of just the access token
* Extend the OAuthTokenResponse with OAuthApiResponse properties. This way, we can parse OAuth errors that come with a different format in comparison with Okta resource APIs
* Update the OktaPagedCollectionEnumerator so the DPoP header is added in paginated results when DPoP is enabled
* Update the ApiClient.ExecuteWithRetryHeaders to generate a new DPoP JWT when it's applicable
* Update templates with the proposed DPoP solution and regenerate the SDK to make sure everything is in place
* Add integration tests to cover DPoP scenarios
* Add unit tests to cover DPoP scenarios
* Update readme with a note about DPoP support
* This feature will be released in version 8.0.0 since it has several breaking changes to make the code clearer

> Note for reviewers: If you want to review the feature right before the templates were modified to reduce the number of files to review, refer to the following [commit](https://github.com/okta/okta-sdk-dotnet/pull/711/commits/8a5f8117a32e335cd14891976a57ab9cd1df2c65)


src/Okta.Sdk/Api/OAuthApi.cs
```diff
@@ -183,10 +195,29 @@ public async Task<ApiResponse<OAuthTokenResponse>> GetBearerTokenWithHttpInfoAsy
...
+                jwtSecurityToken = _jwtGenerator.GenerateSignedJWT();
+                
+                localVarRequestOptions.FormParameters.Remove("client_assertion");
+                localVarRequestOptions.FormParameters.Add("client_assertion", jwtSecurityToken.ToString());
+
+                localVarRequestOptions.HeaderParameters["DPoP"].Clear();
+                localVarRequestOptions.HeaderParameters["DPoP"].Add(dpopJwt);
+
+                localVarResponse = await this.AsynchronousClient.PostAsync<OAuthTokenResponse>(accessTokenUri.Trim(), localVarRequestOptions, this.Configuration, cancellationToken).ConfigureAwait(false);
+            }
```
bryanapellanes-okta, line 219: The new logic here should be delegated to the JwtGenerator implementation.

src/Okta.Sdk/Api/UserApi.cs
```diff
@@ -2415,8 +2416,15 @@ public async System.Threading.Tasks.Task<Okta.Sdk.Client.ApiResponse<User>> GetU
...
-                localVarRequestOptions.HeaderParameters.Add("Authorization", "Bearer " + token);
+                var tokenResponse = await _oAuthTokenProvider.GetAccessTokenResponseAsync(cancellationToken: cancellationToken);
+                localVarRequestOptions.HeaderParameters.Add("Authorization", $"{tokenResponse.TokenType} {tokenResponse.AccessToken}");
+
+                if (tokenResponse.IsDpopBound)
+                {
+                    var requestAbsoluteUri = new Uri(new Uri(this.Configuration.OktaDomain, UriKind.Absolute), new Uri($"/api/v1/users/{userId}", UriKind.Relative));
+                    var dPopProofJwt = _oAuthTokenProvider.GetDPopProofJwt(accessToken: tokenResponse.AccessToken, uri: requestAbsoluteUri.ToString(), httpMethod: "GET");
+                    localVarRequestOptions.HeaderParameters.Add("DPoP", dPopProofJwt);
+                }
```
bryanapellanes-okta, line 2427: This logic should be delegated to the oAuthTokenProvider.

laura-rodriguez: @bryanapellanes-okta I followed the existing pattern provided by the codegen tooling, where headers and any request params are set by the corresponding APIs. If you take a look at other Api classes you will see this pattern.

src/Okta.Sdk/Api/OAuthApi.cs
```diff
@@ -183,10 +195,31 @@ public async Task<ApiResponse<OAuthTokenResponse>> GetBearerTokenWithHttpInfoAsy
...
+                    jwtSecurityToken = _jwtGenerator.GenerateSignedJWT();
+                    
+                    localVarRequestOptions.FormParameters.Remove("client_assertion");
+                    localVarRequestOptions.FormParameters.Add("client_assertion", jwtSecurityToken.ToString());
+
+                    localVarRequestOptions.HeaderParameters["DPoP"].Clear();
+                    localVarRequestOptions.HeaderParameters["DPoP"].Add(dpopJwt);
+
+                    localVarResponse = await this.AsynchronousClient.PostAsync<OAuthTokenResponse>(accessTokenUri.Trim(), localVarRequestOptions, this.Configuration, cancellationToken).ConfigureAwait(false);
+                }
```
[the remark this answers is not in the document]
laura-rodriguez: Both interfaces require slightly different things, for example, there DPoP interface requires a `RotateKeys` method because the client is responsible for keys, whereas the other JWT interface is for OAuth tokens where the consumer is responsible for keys and not RotateKeys method is required. This is why I decided to create a separate interface.

src/Okta.Sdk/Api/OAuthApi.cs
```diff
@@ -168,9 +171,18 @@ public async Task<ApiResponse<OAuthTokenResponse>> GetBearerTokenWithHttpInfoAsy
...
-            var accessTokenUri = $@"/oauth2/v1/token?grant_type=client_credentials&scope={scopes}&client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer&client_assertion={jwtSecurityToken.ToString()}";
+            var accessTokenUri = "/oauth2/v1/token";
+            //?grant_type=client_credentials
+            //&scope={scopes}
+            //&client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer
+            //&client_assertion={jwtSecurityToken.ToString()}";
+            localVarRequestOptions.FormParameters.Add("grant_type", "client_credentials");
+            localVarRequestOptions.FormParameters.Add("scope", scopes);
+            localVarRequestOptions.FormParameters.Add("client_assertion_type", "urn:ietf:params:oauth:client-assertion-type:jwt-bearer");
+            localVarRequestOptions.FormParameters.Add("client_assertion", jwtSecurityToken.ToString());
```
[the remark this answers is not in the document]
laura-rodriguez: See my comments below.

src/Okta.Sdk/Api/OAuthApi.cs
laura-rodriguez: See my comments below.

src/Okta.Sdk/Api/OAuthApi.cs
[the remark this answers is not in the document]
bryanapellanes-okta: The names of the interfaces imply a common surface, the details you describe should be expressed at the implementation level.

src/Okta.Sdk/Api/UserApi.cs
bryanapellanes-okta: I'm suggesting encapsulation into a class responsible for these details.  Patterns established by codegen tools aren't always best practice.

laura-rodriguez: I couldn't agree more that the patterns established by code-gen tools aren't always the best practice, but there's a side that's not visible here, and it's templates. Templates determine in some way how you do things with this design. APIs are responsible for setting request details such as headers, payload, URIs, etc., So, keeping that in mind, the OAuthTokenProvider is responsible for providing tokens, and the Apis will update the request accordingly.

That being said, let's meet and continue this discussion so we speed up the review process. I'll ping you later.

src/Okta.Sdk/Api/OAuthApi.cs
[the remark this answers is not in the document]
laura-rodriguez: Yeah, I actually had the same feeling at the beginning but decided to create a new interface to reduce the number of breaking changes for the existing interface consumers. I'll revisit it, and let's talk about this too.

duytiennguyen-okta:
I have a question because I have to do this for golang: How do you update caching for the access token?

laura-rodriguez:
@bryanapellanes-okta I've updated my PR based on today's discussion. Feel free to review it at your convenience and resolve open comments if applicable. Thank you for the feedback!

bryanapellanes-okta (dismissed):
Our conversation about the feature clarified the nature of the change and intent of the implementation.  My concerns have been addressed.

arvindkrishnakumar-okta (approved):
LGTM 🚢

duytiennguyen-okta (approved):

PR #735, opened 2024-08-05
OKTA-789927: Move away from orb

rlepage-okta:
This moves away from orb-defined jobs when running static analysis tooling.


bryanapellanes-okta (approved):
yippy-kai-yay

PR #739, opened 2024-08-22
v9.0.0

### LINKED ISSUE #650: Factors Api doesn't return factors setup on users
### Describe the bug?

Setting multifactors on a user in the Okta api, and attempting to retrieve them using the latest Okta SDK doesn't return any values.

The raw api response when calling through postman returns the expected values.
`/api/v1/users/:userId/factors`

I should get a list of factors I've configured for the user (4)

0 are returned to me from the sdk api call I make with `IUserFactorApiAsync`

### Reproduction Steps?

1. Create a user in the okta admin
2. Activate the user
3. Login to the user 
4. Setup a multifactor (I setup 4 on my user) on the user account.
5. Use Okta SDK to query the factor Api to list the factors with the okta ID

```
var factors = _factorApi.ListFactors("{{oktaId}}", default(CancellationToken));
var factorList = await factors.ToListAsync();
```

### Additional Information?

_No response_

### .NET Version

6.0.408

### SDK Version

6.0.11
& All versions of 6

### OS version

```
BuildNumber  Caption                             OSArchitecture          Version
19044        Microsoft Windows 10 Enterprise      64-bit               10.0.19044
```

### LINKED ISSUE #718: AppAndInstanceConditionEvaluatorAppOrInstance ID property setter is private
### Describe the bug?

When creating an IdpDiscoveryPolicyRule and attempting to set the rule to apply to a specific application, I am unable to set the Id property in AppAndInstanceConditionEvaluatorAppOrInstance, but it is clearly a valid property that can be set when calling the API as per the documentation at https://developer.okta.com/docs/reference/api/policy/#application-and-app-instance-condition-object.

AppAndInstanceConditionEvaluatorAppOrInstance.Id should be settable.

It is not settable.

### Reproduction Steps?

This is pretty straight forward.  Reproduction isn't necessary.

### Additional Information?

_No response_

### .NET Version

8.0.300

### SDK Version

7.0.6

### OS version

Darwin MacBook-Pro-3.local 23.4.0 Darwin Kernel Version 23.4.0: Fri Mar 15 00:10:42 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6000 arm64

### LINKED ISSUE #719: PolicyRule does not allow a null Priority to be specified.
### Describe the bug?

When creating an IdpDiscoveryPolicy, the API allows a null value to be specified for the rule, in which case the rule will be added just prior to the default rule.   The SDK declares the PolicyRule.Priority property as type int rather than as type int? which prevents passing a null value when creating an IdpDiscoveryPolicyRule.

The value is specifically documented as not being required:
https://developer.okta.com/docs/reference/api/policy/#policy-object

I should be able to not specify a value for Priority and have it send a null value.

A priority of zero is sent when I don't specify a Priority, and an error results since 1 is the minimum allowed value.

### Reproduction Steps?

This is pretty straight forward to identify.

### Additional Information?

_No response_

### .NET Version

8.0.300

### SDK Version

7.0.6

### OS version

Darwin MacBook-Pro-3.local 23.4.0 Darwin Kernel Version 23.4.0: Fri Mar 15 00:10:42 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6000 arm64

bryanapellanes-okta:
- Regenerated from new spec.
- Fixes #650 
- Fixes #718 
- Fixes #719


arvindkrishnakumar-okta (approved):
LGTM 🚢

arvindkrishnakumar-okta (approved):

laura-rodriguez (approved):
🚀

PR #775, opened 2025-02-24
Okta 856169 transform regenerate scripts

bryanapellanes-okta:


bryanapellanes-okta:
merged this with #776

Issue #771, opened 2025-02-04
Factors Api Doesn't return all factors on a user


pdubb29:
### Describe the bug?

When I have a security question factor setup on a user, and I call 
`IUserFactorApiAsync.ListFactors("userId");`
I should be able to get a list of all the factors that are configured on the user.
I've confirmed that when I've setup a Security Question factor on my user, and I call this method, I don't get the user factor that I'm expecting.

I make a call to the user factor api in postman and I get the security question back.
`GET oktaUrl/api/v1/users/:oktaUserId/factors`
Response:
```
[
    {
        "id": "qae2bmppgz4EXeBZT0h8",
        "factorType": "question",
        "provider": "OKTA",
        "vendorName": "OKTA",
        "status": "ACTIVE",
        "created": "2025-02-04T19:13:43.000Z",
        "lastUpdated": "2025-02-04T19:13:43.000Z",
        "profile": {
            "question": "what is this?",
            "questionText": "what is this?"
        },
        ...REDACTED...
        
    },
    {
        "id": "emf25vji1apJTn82V0h8",
        "factorType": "email",
        "provider": "OKTA",
        "vendorName": "OKTA",
        "status": "ACTIVE",
        ...REDACTED....
    }
]
```

### What is expected to happen?

When I call the api in the SDK I should be able to get back the security question from the api call.

### What is the actual behavior?

I get a list of factors that doesn't include my security question

### Reproduction Steps?

1. Allow a user to enroll in a security question and an email multifactor
2. Enroll the user
3. Make a call on the factors api, 
`IUserFactorApiAsync.ListFactors("userId");`

Notice there is only one factor returned, and it's not the security questio

### .NET Version

9.0.102

### SDK Version

9.0.3

### OS version

BuildNumber 
22631                  

Caption
Microsoft Windows 11 Enterprise

OSArchitecture
 64-bit

Version
 10.0.22631

pdubb29:
Apologies, this is the code we were calling and it wasn't working anymore
```
var factors = _factorApi.ListFactors(user.Id);
            var factor = await factors.FirstOrDefaultAsync(x => x.FactorType == UserFactorType.Question);
            if (factor == null)
            {
                return false;
            }
```
When we changed it to this it started working
```
 var factors = _factorApi.ListFactors(user.Id);

 string factorId = string.Empty;
 var asyncEnumerator = factors.GetAsyncEnumerator();
 while(await asyncEnumerator.MoveNextAsync())
 {
[...]

PR #777, opened 2025-03-10
Fix Factors Api Doesn't return all factors on a user

aniket-okta:
This PR resolves the issue(#771) by ensuring all enrolled factors are correctly deserialised and returned.

- Reproduced issue by enrolling Security Question and SMS factors and calling ListFactors, confirming they were missing.
- Identified root cause by running FactorScenarios tests, which failed due to incorrect FactorType mapping.
- Fixed deserialization by removing redefined FactorType properties in affected subclasses (UserFactorSMS, UserFactorSecurityQuestion etc).
- Verified changes by triggering OpenAPI generator to ensure correctness.
- Added an integration test to validate ListFactors returns all enrolled factors correctly.
- Confirmed fix by running tests, which now pass as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 Running changelog of releases since `3.1.1`
 
+## 9.1.2
+### Fixed
+- Fixed deserialization issue in `ListFactors` where Security Question and SMS factors were missing or incorrectly mapped.
+- Ensured `FactorType` is correctly inherited from the base `UserFactor` class. (#771)
+
 ## 9.1.0
 ## Fixed
 - GetUser method now returns `User` instead of `UserGetSingleton`.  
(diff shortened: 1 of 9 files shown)


bryanapellanes-okta (changes requested):
The management.yaml file should be updated to reflect these changes and the generator re-run.  You should be able to delete the FactorType property from each of these objects since that property is defined on the base object:
- https://github.com/okta/okta-sdk-dotnet/blob/OKTA-728925-FixFactorsApi/openapi3/management.yaml#L46327
- https://github.com/okta/okta-sdk-dotnet/blob/OKTA-728925-FixFactorsApi/openapi3/management.yaml#L46450
- https://github.com/okta/okta-sdk-dotnet/blob/OKTA-728925-FixFactorsApi/openapi3/management.yaml#L46472

bryanapellanes-okta (approved):
👍

Issue #896, opened 2026-06-26
IdentityProviderApi.CreateIdentityProviderAsync defaults to a property that cannot be edited via api calls


pdubb29:
### Describe the bug?

Creating an identity provider via the IdentityProviderApi.CreateIdentityProviderAsync defaults to a property that cannot be changed.

This is the request that is used in the Admin UI console

[Idp Creation.json](https://github.com/user-attachments/files/29396403/Idp.Creation.json)

This is the default creation with dummy data. I have the ability to edit the request object in the admin UI with unchecking the checkbox for "Use Persistent Name ID (Higher Security)"

In the okta sdk, that property "protocol.settings.honorPersistentNameId" is not a property we're allowed to set.

Therefore, programmatic creation when we want that property set to `false` is not available to us.
In v9.2.3.0, there is no property for "HonorPersistentNameId"

I'm not sure what the new name of this class would be in v10. I haven't upgraded to v10, due to issue #872 & #871 

### What is expected to happen?

As a developer, I should be able to have all options in the UI available to me via properties on the class.

### What is the actual behavior?

There is no ability to edit that property. The api by default creates an idp with honorPersistentNames being set to true.

### Reproduction Steps?

Using the okta sdk, create an `IdentityProvider`
When you create the `Protocol` object on the `IdentityProvider` class, the sdk doesn't give you the ability to set the boolean property `HonorPersistentNameId`

Creating the idp results in that checkbox being checked on all idps being created.

### .NET Version

10.0.301

### SDK Version

9.2.3

aniket-okta:
Fixed in #898 and will be released in v10.1.0 soon.

Issue #875, opened 2026-03-13
UserApi.ListUsers possible race conditions in paging with JWK authentication


MichaelGrafnetter:
### Describe the bug?

`UserApi.ListUsers` randomly fails enumerating users from one of the subsequent pages JWK authentication.
`ApiException` with error 403 (Forbidden) is thrown when fetching one of the pages. The exception is thrown by `MoveNextAsync()` and contains errCode E0000005 and errorSummary "Invalid session". Looks like a race condition to me.

### What is expected to happen?

Fetch all users with no exception. In case of an error, the retry policy should be applied instead of throwing an exception immediately.

### What is the actual behavior?

In some organizations, the `MoveNextAsync()` iterator randomly throws an exception around pages 5-6 when a JWK of API service is used. Retry policy does not get applied and the failed request is only sent once.

Everything works when SSWS API tokens are used.

### Reproduction Steps?

```cs
try
{
    UserApi userApi = new(_oktaConfig);
    int userCount = 0;
    await foreach (var user in userApi.ListUsers(cancellationToken: cancellationToken).ConfigureAwait(false))
    {
        userCount++;
    }
    Console.WriteLine($"Fetched {userCount} Okta users.");
}
catch (ApiException e)
{
    var status = (HttpStatusCode)e.ErrorCode;
    Console.WriteLine($"Error {e.ErrorCode} ({status}) received while trying to fetch Okta users.");
}
```

### .NET Version

10.0.104

### SDK Version

10.0.3

### OS version

Windows 11

aniket-okta:
Fixed in #898 and will be released in v10.1.0 soon — please reopen if you still hit this.

Issue #874, opened 2026-03-13
ListFactors(userId) does not return any factors for any user


MortenAndresen:
### Describe the bug?

The ListFactors(userId) does not return any factors for any users and it seems to be caused by a deserialization problem when the Okta API returns '_links.resend` as a JSON array. I see this when trying the ListFactorsWithHttpInfoAsync(userId) which fails with ...

Cannot deserialize the current JSON array (e.g. [1,2,3]) into type
'Okta.Sdk.Model.LinksResendResend' because the type requires a JSON object.
Path '_links.resend'.

Example raw response from:
GET /api/v1/users/{userId}/factors

{
  "id": "...",
  "factorType": "email",
  "status": "ACTIVE",
  "_links": {
    "resend": [
      {
        "name": "email",
        "href": "...",
        "hints": {
          "allow": ["POST"]
        }
      }
    ]
  }
}

The SDK model ...
    Okta.Sdk.Model.LinksResendResend 
... expects a single object.

### What is expected to happen?

correctly deserialize the output and include the user factors entries in the list.

### What is the actual behavior?

ListFactors() should handle both ...
    _links.resend : { ... }
... and ...
    _links.resend : [ ... ]

### Reproduction Steps?

var config = new Configuration
{
    OktaDomain = "...",
    Token = "...",
    ConnectionTimeout = 30000,
    RequestTimeout = 60000
};

var factorApi = new UserFactorApi(config);

var factors = await factorApi.ListFactors(userId).ToListAsync();

### Additional Information?

* Raw REST call GET /api/v1/users/{userId}/factors returns correct data.
* Older versions of our integration handle the same and other tenant response without issue.
* I have tried with all v10 releases and with the same result for all.
* The Okta Java SDK changelog mentions a fix for a similar issue ...
        "LinksResend.resend array type issue causing MismatchedInputException"
   Which suggests the API may legitimately return an array for this field.

### .NET Version

Runtime: .NET Framework 4.7.2
SDK target: .NET Standard 2.0

### SDK Version

Okta.Sdk version: 10.0.3

### OS version

BuildNumber  Caption                                 OSArchitecture  Version
20348        Microsoft Windows Server 2022 Standard  64-bit          10.0.20348

aniket-okta:
Fixed in #898 and will be released in v10.1.0 soon.

Issue #873, opened 2026-03-11
Unable to retrieve data properly when calling `SchemaApi.GetGroupSchemaAsync`.


Administrator1577:
### Describe the bug?

After creating multiple attributes data in an Okta group, the result obtained by calling `SchemaApi.GetGroupSchemaAsync` with the SDK is null.

### What is expected to happen?

Able to retrieve group attributes data normally

### What is the actual behavior?

Data can be retrieved from the REST API, but the Okta SDK cannot deserialize it correctly.

### Reproduction Steps?

1. creating multiple attributes data in an Okta group
2. the result obtained by calling `SchemaApi.GetGroupSchemaAsync` with the SDK is null

### Additional Information?

This is the result obtained from the REST API：
`{
  "id": "[https://trial-*.okta.com/meta/schemas/group/default"](https://trial-*.okta.com/meta/schemas/group/default%22),
  "$schema": "[http://json-schema.org/draft-04/schema#"](http://json-schema.org/draft-04/schema#%22),
  "name": "group",
  "title": "Okta group",
  "description": "Okta group profile template",
  "lastUpdated": "2026-03-09T03:03:41.000Z",
  "created": "2026-03-03T02:46:38.000Z",
  "definitions": {
    "custom": {
      "id": "#custom",
      "type": "object",
      "properties": {
        "delia_string": {
          "title": "Delia string",
          "description": "attribute group string",
          "type": "string",
          "mutability": "READ_WRITE",
          "scope": "NONE",
          "minLength": 1,
          "maxLength": 100,
          "permissions": [
            {
              "principal": "SELF",
              "action": "READ_WRITE"
            }
          ],
          "master": {
            "type": "PROFILE_MASTER"
          }
        },
        "delia_integer": {
          "title": "Delia integer",
          "description": "attribute integer",
          "type": "integer",
          "required": true,
          "unique": "UNIQUE_VALIDATED",
          "mutability": "READ_WRITE",
          "scope": "NONE",
          "permissions": [
            {
              "principal": "SELF",
              "action": "READ_WRITE"
            }
          ],
          "master": {
            "type": "PROFILE_MASTER"
          }
        },
        "Delia_number_array": {
          "title": "Delia number array",
          "description": "attribute number array",
          "type": "array",
          "required": true,
          "mutability": "READ_WRITE",
          "scope": "NONE",
          "items": {
            "type": "number"
          },
          "permissions": [
            {
              "principal": "SELF",
[...]

PR #898, opened 2026-09-01
fix: serialization, paging, and DPoP bug fixes

aniket-okta:
Groups seven reported bugs by root cause. All changes are made in the OpenAPI templates, with the SDK regenerated via `.okta/regenerate.sh`.

**Serialization / deserialization**
- #871 — `ApplicationApi.ReplaceApplicationAsync` / `CreateApplicationAsync` returned an Internal Server Error
- #872 — `IdentityProviderApi.ReplaceIdentityProviderAsync` / `CreateIdentityProviderAsync` returned an Internal Server Error
- #873 — `SchemaApi.GetGroupSchemaAsync` did not bind the response correctly
- #874 — `ListFactors(userId)` returned no factors
- #896 — `CreateIdentityProviderAsync` defaulted a property that cannot be edited via the API

**Paging**
- #862 — added auto-paging `IAsyncEnumerable` helpers for wrapped-collection Role/Governance list methods
- #875 — the DPoP proof `htu` claim carried the query string, so every paged enumeration failed after the first page with `invalid_dpop_proof`. Paging follows the absolute `next` link from the `Link` header, which includes a query string, while RFC 9449 requires `htu` to exclude it.

Verified against a live preview org: paged enumeration of 28 users now completes over 28 pages with a DPoP-bound API-service app, with SSWS and non-DPoP JWK auth unaffected. Unit tests: 2806 passing, including new coverage for `htu` normalization.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dhiwakar-okta (approved):
LGTM 🚀

Issue #899, opened 2026-09-02
Passing a Configuration instance to a client discards okta.yaml / appsettings.json values


aniket-okta:
Passing a freshly constructed `Configuration` into an API client silently overrides values loaded from `okta.yaml` / `appsettings.json` with the `Configuration` constructor's own defaults.

### Reproduction

`./okta.yaml`:
```yaml
okta:
  client:
    oktaDomain: https://from-yaml.oktapreview.com
    authorizationMode: BearerToken
    clientId: client-id-from-yaml
    connectionTimeout: 45
```

```csharp
// A: file only — correct
var a = Configuration.GetConfigurationOrDefault();
// OktaDomain=https://from-yaml.oktapreview.com  AuthorizationMode=BearerToken  ConnectionTimeout=45

// B: override a single property in code — file values are lost
var b = Configuration.GetConfigurationOrDefault(new Configuration { AccessToken = "tok" });
// OktaDomain=https://subdomain.okta.com  AuthorizationMode=SSWS  ConnectionTimeout=30000
```

Only `clientId` survives in case B, because `Configuration` has no default for it. `oktaDomain`, `authorizationMode` and `connectionTimeout` are replaced by the constructor defaults. The same happens implicitly via `new UserApi(new Configuration { ... })`, since the client constructor calls `GetConfigurationOrDefault(configuration)`.

### Cause

`GetConfigurationOrDefault` layers the caller's object last:

```csharp
.AddYamlFile(homeOktaYamlLocation, optional: true)
.AddJsonFile(applicationAppSettingsLocation, optional: true)
...
.AddObject(configuration, root: "okta:client")
.AddObject(configuration, root: "okta:testing")
.AddObject(configuration);
```

In `Microsoft.Extensions.Configuration` the last provider wins, so every non-null property on the passed object overrides the file — including the values the caller never set, which the constructor populated (`AuthorizationMode = SSWS` at `Configuration.cs:437/476/497`, plus the default `oktaDomain` and `connectionTimeout`).

### Properties affected

Every property the constructor gives a value to, since the object provider serializes it and the file cannot win:

| Property | Constructor default |
| --- | --- |
| `OktaDomain` | `https://subdomain.okta.com` |
| `AuthorizationMode` | `SSWS` |
| `ConnectionTimeout` | `30000` |
| `RequestTimeout` | `0` |
| `MaxRetries` | `2` |
| `Timeout` | `100000` |
| `UseProxy` | `true` |
[...]

PR #883, opened 2026-05-17
fix: restore Policy as request body for createPolicy/replacePolicy

aniket-okta:
## Problem
In v10, the `createPolicy` and `replacePolicy` operations were changed to accept `CreateOrUpdatePolicy` as the request body type. Since `PasswordPolicy`, `OktaSignOnPolicy`, and other concrete subtypes extend `Policy` (not `CreateOrUpdatePolicy`), it became impossible to pass them to `CreatePolicyAsync`/`ReplacePolicyAsync`.

## Fix
Restored the request body schema for `createPolicy` and `replacePolicy` to `Policy` (matching v9 behaviour), and regenerated the affected SDK surface:
- `CreatePolicyAsync(Policy policy, ...)`
- `ReplacePolicyAsync(string policyId, Policy policy, ...)`

Updated call sites in unit and integration tests that were using `CreateOrUpdatePolicy` directly.

## Validation
- ✅ 2776/2776 unit tests pass 
- ✅ PoC: `PasswordPolicy` with complexity settings round-trips correctly against a live org
- ✅ 11/11 Policy API integration tests pass

Issue #882, opened 2026-05-08
Migration guides are AI slop


jzebedee:
### Describe the bug?

The MIGRATING.md and MIGRATION_GUIDE_v10.0.0.md are AI-generated slop containing many factual errors that don't match the real API surface.

### Hallucinated names

The guides mentions these APIs but they either don't exist in v10, don't match the signature claimed in the guide, or they never existed at all (neither in v9 or v10):

* `ClearUserSessions`: wrong name; v10 uses a revocation-style method name
* `ListUserSessions`: wrong name/shape in the guide
* `GetOrgCommunicationSettingsAsync`: doesn't exist
* `UpdateOrgContactUserAsync`: doesn't exist
* `AddGroupTargetToGroupRoleAsync`: wrong role-target method name
* `ListCustomRoles`: wrong custom-role listing method name
* `UpdateUserRiskAsync`: doesn't exist

### "Unchanged" types

The guides mark these APIs as unchanged or functionally identical, but the v10 signatures don't match:

* ApplicationConnectionsApi
* ApplicationSSOApi
* AuthenticatorApi
* AuthorizationServerKeysApi
* GroupApi
* PolicyApi
* SSFTransmitterApi
* SystemLogApi
* UserApi
* UserFactorApi

### What is expected to happen?

Migration guides are part of your responsibility as a library author to accurately represent breaking changes and inform users of changes to the API shape. If @aniket-okta is not going to maintain any real documentation on the API compatibility, then say so instead of releasing a facade.

### What is the actual behavior?

Migration guides are untrustworthy and not authored or signed off by a developer with knowledge of the changes.

### Reproduction Steps?

Attempt to follow the migration guide.

### .NET Version

Any

### SDK Version

v9.2.3 to v10+

aniket-okta:
[...]

Issue #872, opened 2026-03-11
Calling `IdentityProviderApi.ReplaceIdentityProviderAsync` and `IdentityProviderApi.CreateIdentityProviderAsync` returns an 'Internal Server Error'


Administrator1577:
### Describe the bug?

Calling `IdentityProviderApi.ReplaceIdentityProviderAsync` and `IdentityProviderApi.CreateIdentityProviderAsync` methods to update or create an identity provider data results in an 'Internal Server Error'.

### What is expected to happen?

The above methods can be used to successfully create or update an identity provider data.

### What is the actual behavior?

Calling `IdentityProviderApi.ReplaceIdentityProviderAsync` and `IdentityProviderApi.CreateIdentityProviderAsync` methods to update or create an identity provider data results in an 'Internal Server Error'.

### Reproduction Steps?

1. trying to create an identity provider data
2. calling ApplicationApi.ReplaceApplicationAsync to update the application
3. an internal service error was returned.

### .NET Version

.net 8.0

### SDK Version

10.0.3

pdubb29:
Thanks for raising this issue. One of my products relies on this heavily, and I found this which stopped me from upgrading to v10.

pdubb29:
Any update on this issue? When I upgraded to v10.x of this library, I got a bunch of compiler errors.
Examples:
* 'IdentityProviderProtocol' does not contain a definition for 'Endpoints' and no accessible extension method 'Endpoints' accepting a first argument of type 'IdentityProviderProtocol' could be found (are you missing a using directive or an assembly reference?)
* 'IdentityProviderProtocol' does not contain a definition for 'Issuer' and no accessible extension method 'Issuer' accepting a first argument of type 'IdentityProviderProtocol' could be found (are you missing a using directive or an assembly reference?)
* 'IdentityProviderProtocol' does not contain a definition for 'Credentials' and no accessible extension method 'Credentials' accepting a first argument of type 'IdentityProviderProtocol' could be found (are you missing a using directive or an assembly reference?)

aniket-okta:
Fixed in #898 and will be released in v10.1.0 soon.

PR #900, opened 2026-09-02
Release 10.1.0

### LINKED ISSUE #897: Add search parameter support to IGroupApi.ListGroupUsers

The method `ListGroupUsers` in de Okta.SDK doen't have a search parameter. The API behind it [List all member users](https://developer.okta.com/docs/api/openapi/okta-management/management/tags/group/other/listgroupusers) does have this as a querystring option available. 

We would like to have the `search` query string parameter available in the SDK method `ListGroupUsers` of the `IGroupApi`.

IGroupApi.ListGroupUsers
IGroupApiAsync.ListGroupUsers

https://developer.okta.com/docs/api/openapi/okta-management/management/tags/group/other/listgroupusers

https://github.com/okta/okta-sdk-dotnet/blob/master/docs/GroupApi.md#listgroupusers

https://github.com/okta/okta-sdk-dotnet/blob/6dd7f061efa79a6c2c68e8628dbc7142df40d3ad/src/Okta.Sdk/Api/GroupApi.cs#L877

### Additional Information?

In the previous version of the Okta.SDK the search parameter was available. Since version 10 it has been removed.

aniket-okta:
Cuts **10.1.0**, and carries five community-reported fixes of its own.

The release itself is `chore: Bump version to 10.1.0`: `packageVersion` in `openapi3/config.json`, `<Version>` in `Okta.Sdk.csproj`, the regenerated `Configuration.cs` and `API_README.md`, and a `CHANGELOG.md` entry covering both what is in this PR and everything that landed on `master` since 10.0.3 (#875, #896, #874, #873, #872, #871, #882, and the `createPolicy`/`replacePolicy` request body restored in #883).

Fixes #897, #795, #863, #899, #864.

Each fix is one commit, made in `openapi3/management.yaml` or the mustache templates and regenerated - no hand edits to generated code - with unit tests, and verified against a live org.

### `GroupApi.ListGroupUsers` is missing the `search` query parameter (#897)

The operation took no `search`, so filtering group members meant fetching every one of them and filtering client-side. Added the parameter to the spec.

### Paged list errors name the enumerator class instead of the failing endpoint (#795)

`ApiException` from a paged list read `Error calling OktaPagedCollectionEnumerator\`1:` with nothing after the colon, because authorization failures come back with an empty body. The enumerator now names the endpoint it was asked for, and the exception factory falls back to the `WWW-Authenticate` header when the body is empty, which is where Okta names the missing scope:

```
Error calling /api/v1/users: Bearer authorization_uri="https://***/oauth2/v1/authorize", ..., scope="okta.users.read.self", error="invalid_token", ...
```

The type stays `ApiException` rather than becoming `UnauthorizedAccessException`, which would break every existing `catch (ApiException)` around a list call.

### Caller-supplied configuration file, and only explicitly set values overriding it (#863, #899)

Two halves of the same method:

- `GetConfigurationOrDefault(Configuration, string)` loads a configuration file the caller names, YAML or JSON, relative or absolute. Unlike the conventional locations it is required to exist, because it was asked for by name - a missing file throws `FileNotFoundException` rather than quietly talking to whichever org the ambient `okta.yaml` points at. The one-argument overload is unchanged, so this is source-compatible.
- The caller's object no longer overrides the files with its own constructor defaults (#899). `AddObject` serialized every non-null property, so `new Configuration { AccessToken = "..." }` also silently forced `OktaDomain`, `AuthorizationMode`, `ConnectionTimeout`, `MaxRetries`, `RequestTimeout`, `Timeout`, `UseProxy`, `DisableHttpsCheck`, `DateTimeFormat` and `TempFolderPath` back to their defaults. Only values that differ from a pristine `Configuration` are layered on top now.

`AuthorizationMode` is exempt from that comparison and tracks its own assignment, because `SSWS` is both the default and a deliberate credential choice - dropping it would silently switch the caller to the configuration file's mode.

A named file also has to outrank the `OKTA_*` environment variables, which is what broke the build here: the agent has org credentials in its environment, and those were overriding the per-org files the new tests write. Were it left that way, one stray `OKTA_CLIENT_OKTADOMAIN` would pull every per-org file back to the same org, or pair one org's domain with another org's token - defeating the point of asking for the file by name. The file provider now goes on after the environment, and the configuration tests clear and restore the `OKTA_*` variables so they no longer depend on the machine they run on.

### Signing the client assertion with credentials held outside configuration (#864)

`PrivateKey` had to be a JWK in `okta.yaml` or `appsettings.json`, so the key that authenticates a Super Admin application sat in a file. `Configuration.PrivateKeySigningCredentials` takes a `SigningCredentials` instead, so the key can stay in a TPM (`RSACng` over a Windows KSP), an HSM, a non-exportable `X509Certificate2`, or a cloud key store, and only ever sign. It takes precedence over `PrivateKey`, is `[JsonIgnore]`d so it never reaches a configuration provider, and is carried across `GetConfigurationOrDefault` and `MergeConfigurations` by hand.

TPM access itself is deliberately not implemented: it is platform-specific and this package targets `netstandard2.0`. This is the hook that makes it possible.

### Docs and test isolation

The README documents the named configuration file (with its place in the precedence list), `PrivateKeySigningCredentials`, and the auto-paging role extensions added in #862.

`ApiClientShould`, `OktaConfigurationShould` and `OktaApiClientOptionsShould` all read or change process-wide state - the working directory, where `okta.yaml` is looked for, and the `OKTA_*` variables - and xUnit runs collections in parallel, so one could change it out from under another mid-test. They now share a collection. This was an existing race that the new configuration tests made frequent.

## Testing

- 2836 unit tests pass, and the suite was run repeatedly to confirm the working-directory race is gone. Reverting the provider order fails `PreferTheCallerSuppliedFileOverEnvironmentVariables` and nothing else, so the new test guards the fix and the documented environment-over-conventional-files behaviour is unchanged.
- Reproduced the build failure locally by exporting the same variables the build agent sets, got the same six failures, and confirmed they pass afterwards.
- `.okta/regenerate.sh` produces a tree identical to this branch, so nothing here is a hand edit to generated code.
- Live verification against a preview org for each issue: paged and non-paged error messages, `search` on `ListGroupUsers`, per-org configuration files, and a `private_key_jwt` service application authenticating with a key Okta only ever knew the public half of - including that a key the org does not know is still rejected, and that the existing JWK and SSWS paths are unaffected. Temporary applications and authorization servers created for those runs were deleted.


dhiwakar-okta (approved):
LGTM 🚀

